PR #3, opened 2018-09-27
Molparse

loriab:


qcelemental/molparse/to_schema.py
```diff
@@ -0,0 +1,90 @@
...
+
+import numpy as np
+
+import qcelemental as qcel
+
+from ..util import unnp
+from ..exceptions import *
+from .to_string import formula_generator
+
+def to_schema(molrec, dtype, units='Angstrom', return_type='json'):
```
dgasmith: Needs to return `Bohr` by default to conform to QCSchema.

qcelemental/molparse/to_schema.py
```diff
@@ -0,0 +1,90 @@
...
+    """Translate Psi4 json Molecule spec into json from other schemas.
+
+    Parameters
+    ----------
+    molrec : dict
+        Psi4 json Molecule spec.
+    dtype : {'psi4', 1}
+        Molecule schema format.
+    units : {'Angstrom', 'Bohr'}
```
dgasmith: With conversion factor could be any possible valid unit.

qcelemental/molparse/to_schema.py
```diff
@@ -0,0 +1,90 @@
...
+        # hmm, psi4/qcdb for provenance or does psi molrec need a passthrough field to preserve?
+        #qcschema['provenance'] creator, version, routine
+
+    qcschema = unnp(qcschema)
+
+    if return_type == 'json':
+        return json.dumps(qcschema)
+    elif return_type == 'yaml':
+        import yaml
+        return yaml.dump(qcschema)
```
dgasmith: Why return strings here?

qcelemental/tests/test_molparse_from_string.py
```diff
@@ -0,0 +1,1136 @@
...
+    subject = subject1 + '\n1 1'
+    ans = copy.deepcopy(ans1)
+    ans.update({'fragment_charges': [1.], 'fragment_multiplicities': [1]})
+
+    final, intermed = qcelemental.molparse.from_string(subject, return_processed=True)
+    assert compare_dicts(ans, intermed, 4, sys._getframe().f_code.co_name + ': intermediate')
+    assert compare_molrecs(fullans1c, final['qm'], 4, sys._getframe().f_code.co_name + ': full')
+
+
+def test_psi4_qm_1e():
```
dgasmith: It would be be good to convert these to parametrize at some point.

qcelemental/util/misc.py
```diff
@@ -0,0 +1,108 @@
+import re
+import sys
+import math
+
+import numpy as np
+
+from ..units import conversion_factor
+
+if sys.version_info >= (3, 0):
```
dgasmith: Py3 only for QCelemental I believe?

qcelemental/tests/test_molparse_reconcile_nucleus.py
```diff
@@ -0,0 +1,211 @@
...
+
+co_dominant = (59, 27, 'Co', 58.93319429, True, '')
+co_dominant_mine = (59, 27, 'Co', 58.93319429, True, '_mine')
+co_dominant_shortmass = (59, 27, 'Co', 58.933, True, '')
+co60 = (60, 27, 'Co', 59.93381630, True, '')
+co60ghost = (60, 27, 'Co', 59.93381630, False, '')
+co_unspecified = (-1, 27, 'Co', 60.6, True, '')
+
+
+def test_reconcile_nucleus_1():
```
dgasmith: Great spot for parametrize. Also need to look into the float comparison using `==`.

qcelemental/tests/test_molparse_from_string.py
```diff
@@ -0,0 +1,1136 @@
...
+}
+
+fullans1a = {
+    'geom': np.array([0., 0., 0., 1., 0., 0.]),
+    'elea': np.array([16, 1]),
+    'elez': np.array([8, 1]),
+    'elem': np.array(['O', 'H']),
+    'mass': np.array([15.99491462, 1.00782503]),
+    'real': np.array([True, True]),
+    'elbl': np.array(['', '']),
```
dgasmith: We should probably have a name for these non-schema objects. It is ok to carry around extra metadata, but the fact that units can be something besides Bohr is a dealbreaker.

qcelemental/molparse/to_schema.py
```diff
@@ -0,0 +1,90 @@
...
+    name = molrec.get('name', formula_generator(molrec['elem']))
+#    tagline = """auto-generated by qcdb from molecule {}""".format(name)
+
+    if dtype == 'psi4':
+        qcschema = copy.deepcopy(molrec)
+        qcschema['geom'] = geom
+        qcschema['units'] = units
+        qcschema['name'] = name
+
+    elif dtype == 1:
```
dgasmith: What is a 1 dtype?

qcelemental/molparse/from_string.py
```diff
@@ -0,0 +1,695 @@
...
+            # just a number - must be a CID
+            pcobj = pubchem.PubChemObj(int(pubsearch), '', '')
+            try:
+                xyz = pcobj.getMoleculeString()
+            except Exception as e:
+                raise ValidationError(e.message)
+        else:
+            # search pubchem for the provided string
+            try:
+                results = pubchem.getPubChemResults(pubsearch)
```
dgasmith: I think this function is us, we should change to snake_case at some point.

qcelemental/util/misc.py
loriab: yes, looking forward to a ~Py2~ pass

qcelemental/molparse/to_schema.py
loriab: You wanted simple versions for QCSchema, so this is it at version `1`. `dtype='psi4'` is the expanded dict that (a) has a much larger required fields set, (b) has slightly different systematic field names (elez, elea, elem, etc.), and (c) is the intermediate repr that molparse works with (result of `Molecule.to_dict()`).

And, yeah, QCSchema gets [shut down fast](https://github.com/qcdb/QCElemental/blob/molparse/qcelemental/molparse/to_schema.py#L59-L61) if you try to ask for it in non-Bohr

dgasmith: Oh gotcha, dtype is schema version.

qcelemental/tests/test_molparse_from_string.py
loriab: I've parametrized all the others, but `from_string` is on the messy side. saving it for a refactor/expansion.

qcelemental/molparse/to_schema.py
loriab: good news, `to_schema` is completely unused. it can return whatever's convenient.

qcelemental/tests/test_molparse_from_string.py
dgasmith: The new version looks really clean I think!

PR #8, opened 2018-10-26
fix installed tests

loriab:
forgive comparison tests for installed repo (psi4 and cfour ref data not installed)

### DIFF
--- a/qcelemental/tests/test_constants.py
+++ b/qcelemental/tests/test_constants.py
@@ -78,6 +78,7 @@ def test_c_header():
     os.remove("header.h")
 
 
+@pytest.mark.xfail(True, reason='comparison data not available for installed repository', run=True, strict=False)
 def test_constants_comparison():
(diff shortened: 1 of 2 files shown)


qcelemental/tests/test_constants.py
```diff
@@ -78,6 +78,7 @@ def test_c_header():
     os.remove("header.h")
 
 
+@pytest.mark.xfail(True, reason='comparison data not available for installed repository', run=True, strict=False)
```
dgasmith: Should we check if these files are there or not?

dgasmith: I guess if it is `xfail` its fine. No reason to do anything complex.

loriab: Yeah, it'll xpass if source, xfail if installed. Agree there's no need to make tests a test of where it's running from. The conda pkg is `v0.1.0a` + this PR.

PR #10, opened 2018-10-31
covalent radii

loriab:
Not polished (particularly on elements beyond covrad definitions), but mostly there. No urgency.

- [x] adds `Datum.to_units()` to return data in native or requested units
- [x] adds covalent radii from DaltonTransactions as `qcel.covalentradii`. Natively in Ang from the paper, but request alternate through `qcel.covalentradii.get('Kr', units='bohr')`


qcelemental/covalent_radii.py
```diff
@@ -0,0 +1,148 @@
...
+        text = [
+            '#ifndef _qcelemental_covrad_h_',
+            '#define _qcelemental_covrad_h_',
+            '',
+            '/* This file is autogenerated from the QCElemental python module */',
+            '',
+            'const double covalent_radii[] = {',
+        ]
+
+        for el in periodictable.E[1:50]:
```
loriab: TODO: fix this range

dgasmith: Did you still need to fix this?

qcelemental/covalent_radii.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # add alternate names to help QC programs
+        for alias in aliases:
+            ident, units, value, comment = alias
+            self.cr[ident.capitalize()] = datum.Datum(ident, units, value, comment)
+
+    def __str__(self):
+        return "CovalentRadii(context='{}')".format(self.name)
+
+    def get(self, atom, return_tuple=False, units=None):
```
dgasmith: Would be much happier if default units was `bohr` so that we can bank on this as the default.

loriab: Do you want the units arg here to default to "bohr" instead of "natural"? I'm good with that so long as the internal (in Decimal; read from table) can remain Ang.

This `pint` is going to drag our unit abbreviations into the mainstream. I already learned `Dyne`-->`dyn`.

qcelemental/covalent_radii.py
loriab: Yes. I only saw it as I was packing up, so left myself a note. I think `periodictable` needs an iterator fn.

loriab:
Does the API seem good to you? `Datum.to_units(unit)` and `covalentradii.get(atom, unit)` aren't symmetric but were orderly enough.

qcelemental/covalent_radii.py
dgasmith: Yes please on the `Bohr`. Happy to have internal remain as Decimal and even perhaps the tuple return in Ang as well.  Yes, pint will straighten us out with a singular definition of abbreviations (a good thing).

dgasmith:
`covalentradii.get(atom, units="Bohr")` right? But overall seems fine with me.

PR #13, opened 2019-01-15
enhance unnp fn

loriab:
Recursive ndarray-->list fn now handles lists and flattening options. Requires py3.6 (as a lot more important stuff will shortly).

Nothing pressing on this. Just wanted to get it on GH.

### DIFF
--- a/qcelemental/util/misc.py
+++ b/qcelemental/util/misc.py
@@ -91,22 +91,42 @@ def filter_comments(string):
     return string
 
 
-def unnp(dicary, path=None):
-    """Return `dicary` with any ndarray values replaced by lists."""
-    # TODO handle complex numbers/arrays
+def unnp(dicary, flat=False, _path=None):
+    """Return `dicary` with any ndarray values replaced by lists.
 
-    if path is None:
-        path = []
+    Parameters
+    ----------
+    dicary: dict
+        Dictionary where any internal iterables are dict or list.
+    flat : bool, optional
+        Whether the returned lists are flat or nested.
+
+    Returns
+    -------
+    dict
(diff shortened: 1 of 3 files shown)


qcelemental/util/misc.py
```diff
@@ -91,22 +91,42 @@ def filter_comments(string):
...
-    """Return `dicary` with any ndarray values replaced by lists."""
-    # TODO handle complex numbers/arrays
+def unnp(dicary, flat=False, _path=None):
+    """Return `dicary` with any ndarray values replaced by lists.
 
-    if path is None:
-        path = []
+    Parameters
+    ----------
+    dicary: dict
```
dgasmith: Should flat default to True as this is expected in schema?

loriab:
"pint not found" has stopped two professors and two grad students in their tracks. they may be bold enough to `conda install pint`, but when that fails, they need guidance.

qcelemental/util/misc.py
loriab: defaulting to `flat=True` complicates fragments, since that's properly nested. I keep it as ndarray internally for handy arithmetic.

qcelemental/physical_constants/context.py
```diff
@@ -2,12 +2,15 @@
 Contains relevant physical constants
 """
 
-import collections
-import pint
 from decimal import Decimal
+import collections
 
-from .. import datum
+try:
```
dgasmith: Maybe put this in either the `p/c` init or the main `init` to make sure this is the first instance imported.

loriab: sure. added analogous pydantic warning to #14, too.

PR #14, opened 2019-01-18
Add Molecule Pydantic Object

Lnaden:
First pass for review and design consideration.

Should be functionally complete, but not quite feature or doc complete yet, nor are the tests robust yet. 

I do have a few outstanding design issues to consider:
1. How to handle `required_definitions`?
1. How to properly carry `hash_fields`? Right now its a property, but that requires the instanced class
1. Do we want to cast some objects to more intuitive NumPy arrays (e.g. geometry) or keep them as lists of lists?


qcelemental/tests/test_models.py
```diff
@@ -0,0 +1,44 @@
...
+}
+
+
+def test_molecule():
+    molecule = Molecule(**mol)
+    assert molecule.geometry.flatten().tolist() == mol['geometry']
+    for con_mol, con_raw in zip(molecule.connectivity, mol['connectivity']):
+        assert list(con_mol) == list(con_raw)
+    assert molecule.symbols == mol['symbols']
+    assert all(Molecule.parse_raw(molecule.json()).geometry == molecule.geometry)
```
dgasmith: Make sure to use `np.allclose` or similar here.

qcelemental/tests/test_models.py
```diff
@@ -0,0 +1,44 @@
...
+               0.993691,  1.671622,  5.530409],
+  "provenance": {},
+  "connectivity": [[0, 2, 1], [0, 4, 1], [0, 5, 1], [0, 6, 1], [1, 3, 1], [1, 7, 1], [1, 8, 1],
+                   [1, 9, 1], [2, 3, 1], [2, 10, 1], [2, 11, 1], [3, 12, 1], [3, 13, 1]]
+}
+
+
+def test_molecule():
+    molecule = Molecule(**mol)
+    assert molecule.geometry.flatten().tolist() == mol['geometry']
```
dgasmith: `np.allclose` due to floats.

qcelemental/models/objects/molecule.py
```diff
@@ -0,0 +1,107 @@
...
+    connectivity: List[Tuple[int, int, int]] = []
+    fragments: List[List[int]] = None
+    fragment_charges: List[float] = None
+    fragment_multiplicities: List[int] = None
+    fix_com: bool = False
+    fix_orientation: bool = False
+    provenance: dict = {}
+
+    @validator('geometry')
+    def must_be_3n(cls, v, values, **kwargs):
```
dgasmith: Before we go too deep on this kind of validation with pydantic itself, @loriab do we have validation routines that can be used here?

loriab: for testing, there's https://github.com/MolSSI/QCElemental/blob/master/qcelemental/tests/utils_compare.py#L134 that after `from utils import *` is ready to be used with `assert`.

qcelemental/models/objects/molecule.py
```diff
@@ -0,0 +1,107 @@
...
+class Molecule(BaseModel):
+    id: str = None
+    symbols: List[str]
+    geometry: NPArray = ...
+    masses: List[float] = None
+    name: str = None
+    identifiers: Identifiers = None
+    comment: str = None
+    charge: float = 0.0
+    multiplicity: int = 1
```
loriab: `molecular_charge` and `molecular_multiplicity`

qcelemental/models/objects/molecule.py
loriab: I'd very much recommend stuffing all these parameters into [molparse.from_arrays](https://github.com/MolSSI/QCElemental/blob/master/qcelemental/molparse/from_arrays.py#L130). That'll keep all the validation code localized and also validate self-consistently, rather than field-by-field.

dgasmith: Recommend staying away from _any_ `from utils import *`

loriab: agree in general. this is testing only and a [tight little file](https://github.com/MolSSI/QCElemental/blob/master/qcelemental/tests/utils.py), so i've considered it harmless. now that I look at it, I should privatize the `true_false_decorator`.

loriab:
@Lnaden if the `required_definitions` you mean geom+symbols, I think it's safe to let `from_arrays` take care of it. It has rather looser requirements (https://github.com/MolSSI/QCElemental/blob/master/qcelemental/molparse/from_arrays.py#L168-L169), but the geom+symbols fits in, and you can always get the qcschema with exactly geom+symbols (among other fields) out again.

dgasmith:
I wouldn't worry too much about validation on this pass. This is mostly just about getting the model in so that we can start using them throughout the code. We will also likely want some sort of flag to turn rigorous validation off (and only check fields) as we will need to generate many of these objects quickly.

qcelemental/tests/test_models.py
```diff
@@ -0,0 +1,44 @@
...
+               1.224314, -3.078871,  0.252992,
+               0.687522, -1.875996,  2.724953,
+               5.568104,  1.648263,  6.817593,
+               4.312974,  4.746522,  6.612934,
+               6.29054 ,  3.581917,  4.174546,
+               3.847091, -1.789514,  3.943152,
+               4.559751,  0.050644,  1.254844,
+               1.770032,  3.515675,  2.862177,
+               0.993691,  1.671622,  5.530409],
+  "provenance": {},
```
loriab: btw, whenever you want to generate a provenance, call the [`provenance_stamp`](https://github.com/MolSSI/QCElemental/blob/9196144f4ca4cd7eb0811428e26214a6f6fcf597/qcelemental/molparse/from_string.py#L264). I write one for every program, and it's very handy.

Lnaden:
Still doing work on this one, I think we won't be able to get all the clean validation tools in on this pass, but we can add them on a 2nd round.

There is a bit of issue I am having with how much of a drop-in replcement we want this to be to the existing `QCFractal/interface/molecule.py:Molecule` object which can accept as an `arg` a string or data and build the molecule from that. This new model can only accept an expansion of `kwargs`, so making it a pure drop in will be hard. I can write new constructors, but wanted some opinions first

dgasmith:
I believe we want to:
 - Leave the parsers there for now.
 - Leaving the init as `Molecule(**data)` I think is the right way with `Molecule.from_files` or `Molecule.build(filename/JSON/xyz/etc)`?
 - We will need to rewrite a bit of QCFractal and that is fine.
 - Replace all the parsers with `molparse` code at a later date.

qcelemental/models/common_models.py
```diff
@@ -0,0 +1,10 @@
+from pydantic import BaseModel
+
+
+class Provenance(BaseModel):
+    creator: str
+    version: str = ""
+    routine: str = ""
```
dgasmith: Best to use `Optional[str] = None` here.

qcelemental/models/molecule.py
```diff
@@ -0,0 +1,813 @@
...
+    molecular_charge: float = 0.0
+    molecular_multiplicity: int = 1
+    real: List[bool] = None
+    connectivity: List[Tuple[float, float, float]] = []
+    fragments: List[List[int]] = None
+    fragment_charges: List[float] = None
+    fragment_multiplicities: List[int] = None
+    fix_com: bool = False
+    fix_orientation: bool = False
+    provenance: Provenance = {}
```
dgasmith: Can use the `utils` stamp here.

qcelemental/models/molecule.py
```diff
@@ -0,0 +1,813 @@
...
+                      "to match the schema. They are currently identical in function",
+                      DeprecationWarning)
+        return self.molecular_charge
+
+    @property
+    def multiplicity(self):
+        warnings.warn("This property is being depreciated in favor of `molecular_multiplicity` "
+                      "to match the schema. They are currently identical in function",
+                      DeprecationWarning)
+        return self.molecular_charge
```
dgasmith: We need to detect and nuke these ASAP and get them out of the code. One of the reasons we are building these models.

qcelemental/models/molecule.py
```diff
@@ -0,0 +1,813 @@
...
+        """
+
+        m = hashlib.sha1()
+        concat = ""
+
+        tmp_json = self.to_json()
+        for field in self.hash_fields:
+            if field not in tmp_json:
+                continue
+            concat += json.dumps(tmp_json[field])
```
dgasmith: As a note this will not work without the correct preparation of all of the data (rounding, zero flipping, etc).

qcelemental/models/molecule.py
```diff
@@ -0,0 +1,813 @@
...
+        return m.hexdigest()
+
+    def to_json(self):
+        """
+        Returns a JSON form of the Molecule object.
+        """
+        np.set_printoptions(precision=16)
+        # Because this is expecting a dict of JSON-like objects (not Python),
+        # first cast to the JSON string, then load the json object
+        return json.load(self.json())
```
dgasmith: This is a bit pointless now and likely should just be removed.

qcelemental/models/molecule.py
```diff
@@ -0,0 +1,813 @@
...
+
+    @classmethod
+    def from_json(cls, data, orient=False):
+        # Need to determine how to handle the JSON here
+        if isinstance(data, str):
+            obj = cls.parse_raw(data)
+        else:
+            data['fix_orientation'] = True
+            obj = cls(**data)
+        return obj
```
dgasmith: Ditto here, time to just remove this in favor of Pydantic builders.

Lnaden:
I think this finishes this first implementation of the Molecule class as a 
Pydantic object in QCElemental.

There is one outstanding test that I cannot get to pass due to 1 hash check
failing on the fragments, and I cannot figure how to fix it. The last digit (8th slot) 
in the coordinates is off by a rounding, and the float prep will not 
resolve it.

Lnaden:
I don't know what is wrong with Travis, I cant reproduce the error locally, even trashing all my conda settings.

As for the test which is failing, its in `test_molecule.py:test_water_orient` and specifically `assert frag_0.get_hash() == frag_1.get_hash()` where the two hashes differ in the some of the last digits of the geometries due to rounding differences. Near as I can tell, the code is the same sequence of calls as is done in the original version, but they never quite agree. 

My original thought was the `float_prep` which does `around` operation before and after orient on `__init__` was causing an accumulation of rounding error, which the old `Molecule` class does not do since the geometries are simply assigned after the fragments are extracted. So I re-wrote the `get_fragment` function to make the same `float_prep` calls in the same places, and it still will not align. I have no idea what the fix is here.

loriab:
From where the first error starts on travis, it doesn't seem to like the channel switch-out. Maybe try going back to default channel and nabbing pydantic from pip.

Then again, the travis errors I'm reading don't match your description, so I may be way off base.

Lnaden:
> From where the first error starts on travis, it doesn't seem to like the channel switch-out. Maybe try going back to default channel and nabbing pydantic from pip.

That is exactly what is happening on Travis, which would lead me to think the conda env file is not being read correctly, but all my local tests read the file correctly and don't fail, even after I blow away all of my conda settings. I think the install it from pip for now will work as a temporary fix, its just confusing is all. I'll spend some time with the travis docker image today to figure it out, but I'll set the pip fetch for now

PR #16, opened 2019-01-22
Add the new Results and ResultsInput models to Repo

Lnaden:
Outstanding issues:
* `basis_spec`'s are omitted for now to get this model in. @dgasmith indicates this is probably fine for now
* Missing tests. Do we have a good place to pull some sample data from that isn't me just writing the dicts that also does not have to execute any computation though something like QCEngine?

### DIFF
--- a/qcelemental/models/results.py
+++ b/qcelemental/models/results.py
@@ -64,6 +64,7 @@ class Error(BaseModel):
 
 class ResultsInput(BaseModel):
     """The MolSSI Quantum Chemistry Schema"""
+    id: str = None
     molecule: Molecule
     driver: DriverEnum
     model: Model
--- a/qcelemental/models/results.py
+++ b/qcelemental/models/results.py
@@ -1,0 +1,80 @@
+from enum import Enum
+from pydantic import BaseModel, constr
+from typing import List, Union
+from .molecule import Molecule
+from .common_models import Provenance
+from ..util import provenance_stamp
+
+
+class DriverEnum(str, Enum):
+    energy = 'energy'
+    gradient = 'gradient'
+    hessian = 'hessian'
+
+
+class Model(BaseModel):
+    method: str
+    basis: str = None
+    # basis_spec: BasisSpec = None  # This should be exclusive with basis, but for now will be omitted
+
+
+class Properties(BaseModel):
+    scf_one_electron_energy: float
+    scf_two_electron_energy: float
+    nuclear_repulsion_energy: float
+    scf_vv10_energy: float
+    scf_xc_energy: float
+    scf_dispersion_correction_energy: float
+    scf_dipole_moment: List[float]
+    scf_total_energy: float
+    scf_iterations: int
+    mp2_same_spin_correlation_energy: float
+    mp2_opposite_spin_correlation_energy: float
+    mp2_singles_energy: float
+    mp2_doubles_energy: float
+    mp2_total_correlation_energy: float
+    mp2_total_energy: float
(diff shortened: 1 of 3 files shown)


Lnaden:
Follow up question: Do people see the need to add methods to these classes, or just leave them as effectively lookup models?

loriab:
well, I do have a list (https://github.com/loriab/qcdb/blob/master/qcdb/modelchems.py#L452). but I think leave it free-form.

I'm already having fun with the required basis field for methods like hf-3c, gaussian-2, and sns-mp2. :-) `basis='(auto)'` coming in handy.

qcelemental/models/results.py
```diff
@@ -0,0 +1,84 @@
...
+    mp2_total_energy: float = None
+    calcinfo_nbasis: int = None
+    calcinfo_nmo: int = None
+    calcinfo_nalpha: int = None
+    calcinfo_nbeta: int = None
+    calcinfo_natom: int = None
+    return_energy: float = None
+
+    class Config:
+        allow_extra = False
```
dgasmith: `allow_extra = True` this really was best left as a dictionary field that was not validated at the moment.

qcelemental/models/results.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class ErrorEnum(str, Enum):
+    convergence_error = "convergence_error"
+    file_error = "file_error"
+    memory_error = "memory_error"
```
dgasmith: Not yet strict enum.

Lnaden: The schema makes it a strict enum at the moment, but we could convert it back to a `Dict[str, str]` for now

qcelemental/models/results.py
dgasmith: Eh, I say leave it as is for the moment.

qcelemental/models/results.py
```diff
@@ -0,0 +1,84 @@
...
+
+    class Config:
+        allow_mutation = False
+
+
+class Results(ResultsInput):
+    properties: Properties
+    success: bool
+    error: Error = None
+    return_result: Union[float, List[float]]
```
dgasmith: This needs a `allow_extra=True`. That might overwrite the `Config`, if so the `allow_mutation=False` is good here.

qcelemental/models/results.py
dgasmith: Please do.

qcelemental/models/results.py
Lnaden: Just to confirm, you want the `Result` object to support additional arbitrary fields and still be immutable, correct?

dgasmith:
Is this ready to go?

qcelemental/models/results.py
```diff
@@ -0,0 +1,88 @@
...
+    error_type: Dict[str, str]  # Error enumeration not yet strict
+    error_message: str
+
+    class Config:
+        allow_extra = False
+
+
+### Primary models
+
+class ResultsInput(BaseModel):
```
dgasmith: This should be `Result` singular.

Lnaden: Do you want `Result` singular as well?

Lnaden:
> Is this ready to go?

Barring the plurality check, are there any test data you can point me at to add to the model here before we merge it? Otherwise, its probably ready

Lnaden:
This should be good to go from my perspective. I think as we integrate this into the rest of QCArchive, we will have some decent test data we can add back in

PR #17, opened 2019-01-24
Reworked the Results model to to generalize a few functions and have graceful errors

Lnaden:
As it says on the tin

Added in an Optimization Proceduere for our first procedure action.

There will need to be work to be more general later, but this should be good for now

### DIFF
--- a/qcelemental/models/common_models.py
+++ b/qcelemental/models/common_models.py
@@ -1,5 +1,6 @@
 from pydantic import BaseModel
 from enum import Enum
+from typing import Any
 
 
 class Provenance(BaseModel):
@@ -27,13 +28,24 @@ class DriverEnum(str, Enum):
     hessian = 'hessian'
 
 
-class QCEngineError(BaseModel):
+class ComputeError(BaseModel):
     """The type of error message raised"""
     error_type: str  # Error enumeration not yet strict
     error_message: str
 
     class Config:
         allow_extra = False
+
+
+class FailedOperation(BaseModel):
+    id: str = None
+    input_data: Any = None
+    success: bool = False
+    error: ComputeError
+
+    class Config:
+        allow_extra = True
+        allow_mutation = False
 
 
 qcschema_input_default = "qcschema_input"
--- a/qcelemental/models/common_models.py
+++ b/qcelemental/models/common_models.py
@@ -1,4 +1,5 @@
 from pydantic import BaseModel
+from enum import Enum
 
 
 class Provenance(BaseModel):
@@ -8,3 +9,34 @@ class Provenance(BaseModel):
 
     class Config:
         allow_extra = True
+
+
+class Model(BaseModel):
+    method: str
+    basis: str = None
+    # basis_spec: BasisSpec = None  # This should be exclusive with basis, but for now will be omitted
+
+    class Config:
+        allow_mutation = False
+        allow_extra = True
+
+
+class DriverEnum(str, Enum):
+    energy = 'energy'
+    gradient = 'gradient'
+    hessian = 'hessian'
+
+
+class QCEngineError(BaseModel):
+    """The type of error message raised"""
+    error_type: str  # Error enumeration not yet strict
+    error_message: str
+
+    class Config:
+        allow_extra = False
+
+
+qcschema_input_default = "qcschema_input"
+qcschema_output_default = "qcschema_output"
+qcschema_optimization_input_default = "qcschema_optimization_input"
+qcschema_optimization_output_default = "qcschema_optimization_output"
--- a/qcelemental/models/procedures.py
+++ b/qcelemental/models/procedures.py
@@ -1,0 +1,119 @@
+from pydantic import BaseModel, constr
+from .common_models import Model, DriverEnum, \
+    QCEngineError, \
+    qcschema_input_default, qcschema_optimization_input_default, qcschema_optimization_output_default
+from .molecule import Molecule
+from .results import Result, FailedResult
+from typing import List, Union, Any
+
+
+class InputSpecification(BaseModel):
+    id: str = None
+    driver: DriverEnum
+    schema_name: constr(strip_whitespace=True,
+                        regex=qcschema_input_default) = qcschema_input_default
+    schema_version: int = 1
+    model: Model = None
+    keywords: dict = {}
+
+    class Config:
+        allow_extra = True
+        allow_mutation = False
+
+
+class OptimizationInput(BaseModel):
+    id: str = None
+    schema_name: constr(strip_whitespace=True,
+                        regex=qcschema_optimization_input_default) = qcschema_optimization_input_default
+    schema_version: int = 1
+    keywords: dict = {}
+    input_specification: InputSpecification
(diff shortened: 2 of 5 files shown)


qcelemental/models/common_models.py
```diff
@@ -8,3 +9,34 @@ class Provenance(BaseModel):
...
+        allow_extra = True
+
+
+class DriverEnum(str, Enum):
+    energy = 'energy'
+    gradient = 'gradient'
+    hessian = 'hessian'
+
+
+class QCEngineError(BaseModel):
```
dgasmith: Should just be a `ComputeError` not specific to QCE.

qcelemental/models/procedures.py
```diff
@@ -0,0 +1,119 @@
...
+                        regex=qcschema_optimization_output_default) = qcschema_optimization_output_default
+    final_molecule: Molecule
+    trajectory: List[Result] = None
+    energies: Union[List[float], float] = None
+    success: bool
+    error: QCEngineError = None
+
+
+
+    __json_mapper = {
```
dgasmith: Can we remove everything below this line, should not be here.

qcelemental/models/procedures.py
```diff
@@ -0,0 +1,119 @@
...
+    success: bool = False
+    error: QCEngineError
+
+
+class Optimization(FailedOptimization):
+    schema_name: constr(strip_whitespace=True,
+                        regex=qcschema_optimization_output_default) = qcschema_optimization_output_default
+    final_molecule: Molecule
+    trajectory: List[Result] = None
+    energies: Union[List[float], float] = None
```
dgasmith: `energies` will always be a `List[float]`.

qcelemental/models/procedures.py
```diff
@@ -0,0 +1,119 @@
+from pydantic import BaseModel, constr
+from .common_models import Model, DriverEnum, \
```
dgasmith: Canonical way to do this is with parenthesis `from .common_models import (Model, DriverEnum ...)`.

qcelemental/models/results.py
```diff
@@ -68,21 +47,64 @@ class ResultInput(BaseModel):
...
     schema_version: int = 1
     keywords: dict = {}
     provenance: Provenance = provenance_stamp(__name__)
 
     class Config:
         allow_mutation = False
+        allow_extra = True
+
+
+class FailedResult(ResultInput):
```
dgasmith: I think these Failed classes are a bit much. Best to have a single `ComputeError` class where we have:
```python
success = False
input_data = Any
error_type = str
error_message = str
```

dgasmith:
Thinking through that a bit more in general we will only have that particular return error come up if there is something wrong with the overall JSON structure. By default we will be returning a `Optimization` or `Result` with `success = False` 99.9% of the time with a compute error in its current form.

Perhaps we should leave the ComputeError as is and build out a `UnrecognizedModel` field or similar? Where the creation of the object went wrong.

Lnaden:
> Perhaps we should leave the ComputeError as is and build out a UnrecognizedModel field or similar? Where the creation of the object went wrong.

This is part of the reason why I made the objects in the first place, they keep a pydantic model, but can accept everything, this way, we can treat all the post-compute processing (e.g. the `handle_output_metadata`) the same way since the same object structure will always be returned (Pydantic objects) without being lossy

Lnaden:
I also made the `FailedResult` and `FailedOptimization` objects so that they would all at least have the correct attributes, even if they wound up being `None`

dgasmith:
Then we get an object that is very dangerous, it quacks like a Result but isn't. Anything that fails that hard should be very clear it isn't a valid object.

Lnaden:
> Then we get an object that is very dangerous, it quacks like a Result but isn't. Anything that fails that hard should be very clear it isn't a valid object.

Fair. Would you be okay with something like this:

```python
class ComputeError(BaseModel):
    error_type: str  
    error_message: str

     class Config:
        allow_extra = False

class FailedOperation(BaseModel):
    success: bool = False
    error: ComputeError

    class Config:
        allow_extra = True
        allow_mutable = False

class FailedResult(FailedOperation):
    pass

class FailedOptimization(FailedOperation):
    pass
```

This way they are all common objects, but we can infer where the failure happened by name? Or would you prefer just the `FailedOperation` object?

dgasmith:
There isn't a reason anywhere in the current code where we would need a `FailedResult`. Can you describe a use case for it?

Lnaden:
At the moment, I'm only trapping malformed inputs from `dict` type input, and I skipped over the `except Exception as e:` since we agreed on speed now, refine later. The use case would be if the `get_program(program)(input_data, config)` call in QCEngine failed for whatever reason: program crashes, user provides really dumb inputs which are technically "valid," this would be the catchall.

Lnaden:
The other use case right now is program does not exist, which is possible. I'm fine only having a `FailedOperation` class, but we need some way to pass back a Pydantic object which does no validation except on the error, otherwise we are going to have a ton of additional `try: except` clauses in the code as we try to read back in data and many `if isinstance(obj, dict)` since internally we assume we're going to have the Pydantic models, unless there is an error, then we would need a different set of functional calls since we cant do `{BaseModel}['some_attribute']` and have it work

dgasmith:
Right, which is why I favor `FailedOperation`. But why do we need `FailedResult/FailedOptimization`?

Lnaden:
>  But why do we need FailedResult/FailedOptimization?

We don't, per se, I just was considering making them so we could distingush between Results and Procedure objects since we have no safe/reliable way to separate them unless we inspect the internal attributes and guess

Lnaden:
The counter argument would be results will only ever be expected from certain functions, and procedures from others, so the distinction would be automatic on its own.

dgasmith:
The code in general does not handle failed results at that level. If it worked far enough to get a `Result/Optimization` object with `success = False` we may do something. Everything else is considered not worthwhile.

This result is just one very small piece of puzzle and QCFractal is very aware what each task is. Once the data is turned back into a dict it looses the class description anyway.

Lnaden:
> If it worked far enough to get a Result/Optimization object with success = False we may do something. Everything else is considered not worthwhile.

Kind of, in some cases, it only returned the raw "success/error" dict. So having a "failed" class accept anything with those mandatory would work. 

>  Once the data is turned back into a dict it looses the class description anyway.

Fair. Are you still okay with a result, on any error, returning a `FailedOperation` with the input just fed in as additional attributes (although feeding into a `inputs: input_dict` would be safer). Or do you just want to return a `Failed` object which drops all of the input information?

Lnaden:
I think I have all the requests you made in. Let me know what you think. If you want to make any more changes tonight, feel free to do so as I'll be stepping away for the night barring a disaster (call me if need be)

dgasmith:
Well, definitely will not be calling anyone about code emergencies. We are not SysAdmins ;0

Changes look great, thanks for making them.

Lnaden:
> We are not SysAdmins ;0

Now then, you *say* that....

> thanks for making them.
+:praisethesun:

PR #19, opened 2019-01-25
Docs: Patches build failures by using conda-env/base.yaml for build

dgasmith:
Might be something we should look into for all projects to remove yet another conda env yaml.


readthedocs.yml
```diff
@@ -1,5 +1,5 @@
 conda:
-    file: docs/requirements.yml
+    file: devtools/conda-envs/base.yaml
```
Lnaden: Doesn't doing this (and removing the requirements.yml file) mean that we now need to add all the missing dependencies to the base env file? So sphinx, sphinx_rtd_theme, sphinx-automodapi, and graphviz? I'm not sure we want to do that.

dgasmith: Ah crap, yea.

PR #20, opened 2019-01-27
Molparse

dgasmith:
Moves several molecule functions over to molparse and remove explicit molecule in data folders.


qcelemental/models/molecule.py
```diff
@@ -76,23 +78,38 @@ class Config:
...
     fragment_charges: List[float] = None
     fragment_multiplicities: List[int] = None
+
+    # Orientation
     fix_com: bool = False
     fix_orientation: bool = False
+
+    # Extra
     provenance: Provenance = provenance_stamp(__name__)
+    id: str = None
```
loriab: should all possible fields be listed here? thinking `fix_symmetry`

qcelemental/models/molecule.py
```diff
@@ -177,8 +194,7 @@ def must_be_n_frag(cls, v, values, **kwargs):
                 raise ValueError("Fragment Charges and Fragment Multiplicities"
                                  " must be same number of entries as Fragments")
         else:
-            raise ValueError("Cannot have Fragment Charges or Fragment Multiplicities "
-                             "without Fragments")
+            raise ValueError("Cannot have Fragment Charges or Fragment Multiplicities " "without Fragments")
```
loriab: pertains to `populate_real` [above](https://github.com/MolSSI/QCElemental/pull/20/files#diff-c1df0b88317a0c0331bea59a1793059cR187): so once qcschema_molecule exists, field `real` guaranteed present? this may change intf_dftd3 a smidge

qcelemental/models/molecule.py
```diff
@@ -417,15 +435,15 @@ def get_molecular_formula(self):
     ### Constructors
 
     @classmethod
-    def from_data(cls, data, dtype, orient=False, **kwargs):
+    def from_data(cls, data, dtype=None, *, orient=False, **kwargs):
```
loriab: (aside), i've been thinking we shouldd use `*` in signatures more. for defense.

qcelemental/models/molecule.py
```diff
@@ -438,13 +456,26 @@ def from_data(cls, data, dtype, orient=False, **kwargs):
...
+            elif isinstance(dat, np.array):
+                dtype = "numpy"
+            elif isinstance(data, dict):
+                dtype = "dict"
+            else:
+                raise TypeError("Input type not understood, please supply the 'dtype' kwarg.")
 
-        if dtype == "psi4":
-            input_dict = cls._molecule_from_string_psi4(data)
+        if dtype in ["string", "psi4"]:
```
loriab: if there's a use case where you only wanted to load pure-Cart psi4 (and exclude zmat input), may want to allow psi4/psi4+/xyz/xyz+ in l446 in docstring above.

qcelemental/models/molecule.py
```diff
@@ -438,13 +456,26 @@ def from_data(cls, data, dtype, orient=False, **kwargs):
...
+            input_dict = to_schema(from_string(data)["qm"], 1)["molecule"]
```
loriab: `dtype=1`. 1 is scarily generic

qcelemental/models/molecule.py
```diff
@@ -438,13 +456,26 @@ def from_data(cls, data, dtype, orient=False, **kwargs):
...
         elif dtype == "numpy":
-            input_dict = cls._molecule_from_numpy(data,
-                                                  frags=kwargs.pop("frags", []),
-                                                  units=kwargs.pop("units", "angstrom"))
+            data = {
+                "geom": data[:, 1:],
+                "elez": data[:, 0],
+                "units": kwargs.pop("units", "Angstrom"),
+                "fragment_separators": kwargs.pop("frags", [])
```
loriab: your `frags` definitely the [same as mine](https://github.com/MolSSI/QCElemental/blob/master/qcelemental/molparse/from_arrays.py#L249-L250)?

Reading to end, answer is yes. it's one of your internal flags, not qcschema.

qcelemental/models/molecule.py
loriab: sounds good. in dftd3 I already have to form the molrec for `to_string`, so i'll just keep copying `'real'` until dust settles.

PR #27, opened 2019-02-04
Molecule Measurements

dgasmith:
Adds several routines for computing distances, angles, and dihedrals within QCElemental. In addition adds a `Molecule.measure` command to check internal values.

### DIFF
--- a/qcelemental/models/molecule.py
+++ b/qcelemental/models/molecule.py
@@ -12,7 +12,7 @@ from typing import List, Tuple
 from ..physical_constants import constants
 from ..periodic_table import periodictable
 from .common_models import Provenance, ndarray_encoder
-from ..util import provenance_stamp
+from ..util import provenance_stamp, measure_coordinates
 
 from ..molparse import from_string, from_arrays, to_schema
 
@@ -227,6 +227,13 @@ class Molecule(BaseModel):
             return ret
 
 ### Non-Pydantic API functions
+
+    def measure(self, measurements, degrees=True):
+        """
+        Takes a measurement of the moleucle from the indicies provided.
+        """
+
+        return measure_coordinates(self.geometry, measurements, degrees=degrees)
 
     def orient_molecule(self):
         """
--- a/qcelemental/util/misc.py
+++ b/qcelemental/util/misc.py
@@ -140,6 +140,44 @@ def _norm(points):
     tmp = np.atleast_2d(points)
     return np.sqrt(np.einsum("ij,ij->i", tmp, tmp))
 
+def measure_coordinates(coordinates, measurements, degrees=False):
+    """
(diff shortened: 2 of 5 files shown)


qcelemental/util/misc.py
```diff
@@ -130,3 +130,173 @@ def unnp(dicary, flat=False, _path=None):
...
     return ndicary
+
+
+def _norm(points):
+    """
+    Return the Frobenius norm across axis=-1, NumPy's internal norm is crazy slow (~4x)
+    """
+
+    tmp = np.atleast_2d(points)
+    return np.sqrt(np.einsum("ij,ij->i", tmp, tmp))
```
loriab: not duplicating https://github.com/MolSSI/QCElemental/blob/81c65d815fbb18b8bb689e4a1c4c2b2c896d5c56/qcelemental/util/misc.py#L9-L18, is it?

qcelemental/util/misc.py
```diff
@@ -130,3 +130,173 @@ def unnp(dicary, flat=False, _path=None):
...
+
+def measure_coordinates(coordinates, measurements, degrees=False):
+    """
+    Measures a geometry array based on indices provided, automatically detects distance, angle,
```
loriab: "0-based indices"

qcelemental/util/misc.py
```diff
@@ -130,3 +130,173 @@ def unnp(dicary, flat=False, _path=None):
...
+    Parameters
+    ----------
+    points1 : np.ndarray
+        The first list of points, can be 1D or 2D
+    points2 : np.ndarray
+        The second list of points, can be 1D or 2D
+    points3 : np.ndarray
+        The third list of points, can be 1D or 2D
+    degrees : bool, options
+        Returns the angle in degress rather than radians if True
```
loriab: "degrees" (ditto in dihedral)

qcelemental/models/molecule.py
```diff
@@ -228,6 +228,13 @@ def json(self, *args, **kwargs):
 
 ### Non-Pydantic API functions
 
+    def measure(self, measurements, degrees=True):
```
Lnaden: Are degrees really the default people expect for quantum chemistry applications? Just a side thought is all

qcelemental/util/misc.py
dgasmith: No, this is just a fast norm equivalent to `np.linalg.norm(arr, axis=1)`. Somewhat meaningless, but something I often do.

qcelemental/models/molecule.py
dgasmith: I think so, happy to take other comments here however,

loriab: user-facing is usually degrees, I think. coders would rather convert than deal with confused gen chem students running labs and touting vespr. :-)

PR #28, opened 2019-02-07
Improved import time

dgasmith:
The `qcelemental` load times are ~1 second due to pint which is a bit much for trying things out. Here, I lazy load pint so initial load is now down to 0.4 seconds (pint problems!).

If I additionally remove `molparse` and `models` from the `__init__.py` I move down to 0.2 seconds (effectively the minimum). If we consider this a use case where folks will not be using the models this might be something that we should consider. Any opinions here?

### DIFF
--- a/qcelemental/extras.py
+++ b/qcelemental/extras.py
@@ -1,0 +1,22 @@
+"""
+Misc information and runtime information.
+"""
+
+from . import _version
+
+__all__ = ["get_information"]
+
+versions = _version.get_versions()
+
+__info = {"version": versions['version'], "git_revision": versions['full-revisionid']}
+
+
+def get_information(key):
+    """
+    Obtains a variety of runtime information about QCFractal.
+    """
+    key = key.lower()
+    if key not in __info:
+        raise KeyError("Information key '{}' not understood.".format(key))
+
+    return __info[key]
--- a/qcelemental/extras.py
+++ b/qcelemental/extras.py
@@ -13,7 +13,7 @@ __info = {"version": versions['version'], "git_revision": versions['full-revisio
 
 def get_information(key):
     """
-    Obtains a variety of runtime information about QCFractal.
+    Obtains a variety of runtime information about QCElemental.
     """
     key = key.lower()
     if key not in __info:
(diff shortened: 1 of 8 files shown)


dgasmith:
I think we ever plan to do it, we should do it now. We can include it later *without* breaking code, but if we do `0.4 -> 0.2` later we will break a great deal of code.

loriab:
An excellent point. I don't want to have to do _another_ rearrange driver fundamentals PR to snsmp2. Of the big 4 components -- pt, physconst, molparse, models -- I'm happiest with pt and physconst being the "always on". reasonable to me to remove the latter two if others see that as plausible use case.

dgasmith:
All in favor of popping out models/molparse?

loriab:
can popping out molparse be a v0.3 thing? i'm going to need to change the psi4/qcel pinning again on conda so it can't update to packages with this in it.

dgasmith:
Sure, not in a rush. Merge when ready.

loriab:
this PR in itself is harmless. it's the popping that'll require adjustment. single or separate PRs planned?

dgasmith:
Separate PR would be great. I hope to cut a minor soon.

dgasmith:
Ok, here we are in milliseconds:
```
259 - Total load
 37 - models + pydantic
   9 - molparse
 53 - versioneer (likely less once compiled)
149 - numpy
 11 - misc
```

So 202/259 is out of our hands, we need to NumPy elsewhere anyways and versioneer will likely be much faster without an explicit `git` call. This will let us not bother with the imports.

qcelemental/util/internal.py
```diff
@@ -1,4 +1,4 @@
-from .._version import get_versions
+from qcelemental.extras import get_information
```
loriab: how about `from ..extras import get_information`? it just seems untidy to import whole module from within.

qcelemental/extras.py
```diff
@@ -0,0 +1,22 @@
...
+__all__ = ["get_information"]
+
+versions = _version.get_versions()
+
+__info = {"version": versions['version'], "git_revision": versions['full-revisionid']}
+
+
+def get_information(key):
+    """
+    Obtains a variety of runtime information about QCFractal.
```
loriab: QCElemental

PR #30, opened 2019-02-14
Fixes measure dihedrals

dgasmith:
Out dihedral code was incorrect for some quadrants of a dihedral. This has been fixed with additional tests.

Also adds tests for `json_dict` which incorrect did not have JSON imported.

### DIFF
--- a/qcelemental/util/misc.py
+++ b/qcelemental/util/misc.py
@@ -251,7 +251,7 @@ def compute_angle(points1, points2, points3, degrees=False):
 
 def compute_dihedral(points1, points2, points3, points4, degrees=False):
     """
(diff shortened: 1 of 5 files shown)


qcelemental/util/misc.py
```diff
@@ -276,25 +276,33 @@ def compute_dihedral(points1, points2, points3, points4, degrees=False):
...
-    right = np.einsum("ij,ij->i", n123, n234)
-
-    angle = np.arctan2(left, right)
+    v1 = -1.0 * (points2 - points1)
+    v2 = points3 - points2
+    v3 = points4 - points3
+
+    v2 = v2 / _norm(v2)
+
+    # vector rejections
```
loriab: rejections?

dgasmith: Whoops, fixing.

PR #31, opened 2019-02-14
Minor cleanups for v0.2.5

dgasmith:
iSorts and YAPFs the entire code base + enhancement of tests.


qcelemental/_version.py
```diff
@@ -1,4 +1,3 @@
-
```
loriab: Better to put in a yapf disable around whole file than let a versioneer generates file be edited, right?

docs/source/changelog.rst
```diff
@@ -13,6 +13,19 @@ Changelog
 .. Bug Fixes
 .. +++++++++
 
+0.2.5 / 2019-02-08
+------------------
+
+Enhancements
+++++++++++++
+
+- (:pr:`31`) Lints the code base preparign for a release and minor test improvements.
```
loriab: preparing

qcelemental/data/nist_2014_codata.py
```diff
@@ -2033,4 +2033,4 @@
             'uncertainty': '0.000 0017 e-3'
         }
     }
-}
+} # yapf: disable
```
loriab: 2 sp before comment

qcelemental/tests/test_molecule.py
```diff
@@ -190,13 +235,27 @@ def test_water_orient():
...
+    data["connectivity"] = [(-1, 5, 5)]
+    with pytest.raises(ValidationError):
+        Molecule(**data)
+
+
+def test_molecule_errors_shape():
+    data = water_molecule.dict()
+    data["geometry"] = list(range(8))
+    with pytest.raises(ValidationError):
+        Molecule(**data)
```
loriab: For these very generic errors, if there’s more specific info in the error msg, it’s sometimes handy to test the string contents too.

qcelemental/_version.py
dgasmith: I would say, yes. But since we do not regenerate these, I say just YAPF once and get on with it.

qcelemental/tests/test_molecule.py
dgasmith: Agreed, best to put it on a TODO.

qcelemental/_version.py
loriab: So long as doesn’t interfere with upgrading versioneer, sounds fine. (I don’t think it should, just the one complication that comes to mind)

dgasmith: I don't think versioneer will even be upgraded again to be honest. I think that repository is dead and PyPA is creating a replacement. See [here](https://github.com/pypa/setuptools_scm).

PR #33, opened 2019-02-24
qcschema_molecule version 2

loriab:
- [x] implement internal molecule name/version dtype=2 for `molparse.to_schema` / `molparse.from_schema`
- [x] add schema_name/_version to pydantic molecule model


loriab:
@Lnaden, that seemed too easy that I added two req'd fields to the Molecule model and everything still passed. feel free to let me know or to add more tests.

Lnaden:
It should be fine, you set the defaults and don't add any Validator to enforce `Molecule` dicts be of a certain version, so it should pass just fine. With the addition of our Molecule objects only caching the fields it was given on init, then it should be fine.

loriab:
This is RTG.

I'm not certain this is absolutely needed for psi4 v1.3, so if you feel same, your call on when to mint 0.3.0.

PR #34, opened 2019-02-25
Pydantic Datum class

loriab:
Converts `qcel.Datum` to Pydantic model. Changes
  * (a) comment, doi, glossary fields must be accessed by keyword,
  * (b) `to_dict()` becomes `dict()` and instead of label, units, data only, now comment, doi, glossary present _if_ non-default,
  * (c) complex values no longer list-ified by `to_dict()`.

### DIFF
--- a/qcelemental/datum.py
+++ b/qcelemental/datum.py
@@ -50,10 +50,10 @@ class Datum(BaseModel):
     @validator('data')
     def must_be_numerical(cls, v, values, **kwargs):
         try:
-            ident = 1.0 * v
+            1.0 * v
         except TypeError:
             try:
-                ident = Decimal('1.0') * v
+                Decimal('1.0') * v
             except TypeError:
                 raise ValueError('Datum data should be float, Decimal, or np.ndarray')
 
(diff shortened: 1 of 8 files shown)


qcelemental/datum.py
```diff
@@ -28,8 +29,35 @@ class Datum(collections.namedtuple('Datum', 'label units data comment doi glossa
...
+    def must_be_numerical(cls, v, values, **kwargs):
+        try:
+            ident = 1.0 * v
+        except TypeError:
+            try:
+                ident = Decimal('1.0') * v
+            except TypeError:
+                raise ValueError('Datum data should be float, Decimal, or np.ndarray')
+
+        return v
```
dgasmith: As LGTM mentions ident is never used, did you mean to return this?

loriab: Yes, I had a fix (now pushed). Was waiting to compbine it with any other fixes spotted.

loriab:
I was thinking that the molecule dtype=2 PR and this PR contained new fields and slightly breaking changes such that the next mint that includes them could be `v0.3.0`.

dgasmith:
Sounds good on `v0.3`. I think this is ready to go?

loriab:
yes, I think this and the mol dtype=2 are RTG. If you pull in one, I'll rebase the other and patch up the changelog.

PR #35, opened 2019-02-27
Model updates

dgasmith:
Forbids extras on all base models, and updates fields correspondingly.


qcelemental/models/common_models.py
```diff
@@ -1,55 +1,58 @@
...
 
 ndarray_encoder = {np.ndarray: lambda v: v.flatten().tolist()}
 
 
 class Provenance(BaseModel):
     creator: str
-    version: str = None
-    routine: str = None
+    version: Optional[str] = None
+    routine: Optional[str] = None
```
loriab: this means that all three of creator/version/routine must be present in an instance but latter two may be defaulted if not supplied? molparse is demanding all three be supplied.

qcelemental/models/common_models.py
```diff
@@ -1,55 +1,58 @@
...
 
     class Config:
-        extra = Extra.allow
+        extra = "allow"
 
 
 class Model(BaseModel):
     method: str
-    basis: str = None
+    basis: Optional[str] = None
```
loriab: this'll be helpful for d3

qcelemental/models/common_models.py
dgasmith: Yes, this was always the case. The `Optional` isn't actually used unless a `mypy --strict-optional` is set. In general, version and routine are always set.

loriab: ok. so long as we don't hit cases that pass you and fail me, all's well.

PR #37, opened 2019-03-03
QCA Sphinx Theme

dgasmith:
Updates QCElemental documentation to QCArchive sphinx docs.

Also tweaks ResultProperties so that it is now exposed.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -97,7 +97,11 @@ pygments_style = 'default'
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = 'sphinx_rtd_theme'
+try:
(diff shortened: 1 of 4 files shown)


docs/source/conf.py
```diff
@@ -97,7 +97,11 @@
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = 'sphinx_rtd_theme'
+try:
+    import qcarchive_sphinx_theme
+    html_theme = 'qcarchive_sphinx_theme'
+except ModuleNotFoundError:
+    html_theme = 'sphinx_rtd_theme'
```
loriab: packagers will like this. they get touchy about doing a new package just so they can have docs.

PR #42, opened 2019-03-07
Molecule: Always validate schema and fixes fragment multiplicity issues

dgasmith:
Always runs through the more strict molars validators. Can likely remove pydantic validators here. A problem is that not all issues within molparse raise a QCElemental error type (incorrect array length for one), so backed those down to just check for Exception. This should certainly be a TODO to look at in the future.

PR #44, opened 2019-03-07
from_data to dtype=2

loriab:
(1 commit)

could cause problems downstream

needs rebase after #42 merged


dgasmith:
Can you rebase this for the history?

loriab:
Didn't mean for this to supersede #42, unless you're confident it's harmless to downstream. Which do you want to proceed with?

dgasmith:
It isn't completely harmless downstream as hashes change, but we need it in to start computing I think.

loriab:
ok, rebased. clean pass locally, so clear for rev/merge after CI

PR #46, opened 2019-03-08
Periodic Table Docs

dgasmith:
Adds periodic table docs and tries out another way of documenting APIs.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -47,7 +47,7 @@ or use ``pip`` for a local install:
 
 Testing
 -------
-QCElemental can be tested once install using the ``pytest`` package which can be installed via Conda as well:
+QCElemental can be tested using the ``pytest`` package which can be installed via Conda as well:
 
 .. code-block:: console
 
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -6,9 +6,11 @@ You can install qcelemental with ``conda``, with ``pip``, or by installing from 
 Conda
 -----
 
-You can update qcelemental using `conda <https://www.anaconda.com/download/>`_::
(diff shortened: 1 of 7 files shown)


docs/source/install.rst
```diff
@@ -20,27 +22,47 @@ Pip
...
+or use ``pip`` for a local install:
+
+.. code-block:: console
+
+    >>> pip install -e .
+
+
+Testing
+-------
+QCElemental can be tested once install using the ``pytest`` package which can be installed via Conda as well:
```
loriab: drop "once install"

docs/source/periodic_table.rst
```diff
@@ -0,0 +1,80 @@
...
+    >>> qcel.periodictable.to_mass(6)
+    12.0
+    >>> qcel.periodictable.to_mass("C")
+    12.0
+    >>> qcel.periodictable.to_mass("C12")
+    12.0
+    >>> qcel.periodictable.to_mass("Carbon")
+    12.0
+
+A brevity of infomation can be accessed in this manner. Taking 'Carbon' as an example for mass information:
```
loriab: brevity --> variety?

dgasmith: I was thinking bevy, but variety is better.

loriab: either :-)

docs/source/periodic_table.rst
```diff
@@ -0,0 +1,80 @@
+Periodic Table
+==============
+
+Full access to the periodic table is also available within QCPortal. This
```
loriab: I don't care but is QCPortal what you meant?

dgasmith: Nope. Make a patch?

PR #47, opened 2019-03-10
derint fn and fix_symmetry in Molecule model

loriab:
- [x] psi was failing with for pbeh3c because model didn't have the optional `fix_symmetry` as allowed field
- [x] added a `derint` fn so one can do ready numeric comparisons on `DriverEnum`

### DIFF
--- a/qcelemental/models/common_models.py
+++ b/qcelemental/models/common_models.py
@@ -33,14 +33,12 @@ class DriverEnum(str, Enum):
     hessian = 'hessian'
     properties = 'properties'
 
-    def derint(self):
-        # debatable whether properties should be 0, None, 9, ...
-        if self in ['energy', 'properties']:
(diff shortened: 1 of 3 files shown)


qcelemental/models/common_models.py
```diff
@@ -33,6 +33,15 @@ class DriverEnum(str, Enum):
     hessian = 'hessian'
     properties = 'properties'
 
+    def derint(self):
```
dgasmith: Can we make this name more verbose, and `properties` should return `None`.

loriab: `to_integer` or `derivative_int`?

for `properties` would be nice to have a response that can do math that doesn't require error handling.

dgasmith: Either one on the names. Ok on `properties=0`.

PR #48, opened 2019-03-11
Molecule: Prevent duplicate validates

dgasmith:
Validation takes ~0.04 seconds, this makes it slightly smarter about validation and reduce the total number of calls.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -14,6 +14,31 @@ Changelog
 .. +++++++++
 
 
+0.3.2 / 2019-03-11
+------------------
+
+New Features
+++++++++++++
+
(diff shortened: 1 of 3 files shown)


dgasmith:
@loriab Patched an issue here where ghost fragments carried charges and were therefore validated incorrectly. I don't think the current solution is quite right, please advise.

loriab:
the fragment problem has had a couple approaches. first there was a keyword: https://github.com/MolSSI/QCElemental/blob/master/qcelemental/molparse/chgmult.py#L74-L82 . that was fully implemented, but I ended up not using it with psi, instead do some outright resets https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/molecule.py#L1407-L1434 . but that uses info (Real/Ghost/Absent) not in a qcschema mol.

just more info. haven't reached a recommendation.

PR #49, opened 2019-03-12
molparse profiling pass

loriab:
precompile regexes and recompute massnumber/mass dicts for faster mol validation


dgasmith:
I show a nine-fold speedup here! Awesome, let me look around a bit more.

loriab:
moved the changelog

PR #51, opened 2019-03-21
Molecule: Changes default connectivity to None. 

dgasmith:
`connectivity` defaulted to `[]` instead of None. This doesn't make much sense as `[]` indicates that there are no connections inside the molecule, while `None` should indicate that it simply was not supplied.

WARNING: This will change all molecule hashes that do not already include connectivity.

### DIFF
--- a/qcelemental/models/molecule.py
+++ b/qcelemental/models/molecule.py
@@ -103,7 +103,7 @@ class Molecule(BaseModel):
     mass_numbers: List[int] = None
 
     # Fragment and connection data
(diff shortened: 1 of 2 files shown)


dgasmith:
It just means we need to remake all Molecule hashes when we upgrade elemental.

dgasmith:
@loriab Can you review this one too? Just because it is so fundamental.

dgasmith:
Definitely yes. `[]` is a valid connectivity table if odd. A helium complex for example would fall under this category. `None` means that no connectivity was supplied.

PR #54, opened 2019-04-11
ResultProperties: Added keywords for CCSD quantities.

sjrl:
Added keywords for CCSD quantities. Is this all that needs to be done? Or does the qc_schema also need to be updated?

### DIFF
--- a/qcelemental/models/results.py
+++ b/qcelemental/models/results.py
@@ -56,18 +56,6 @@ class ResultProperties(BaseModel):
     ccsd_prt_pr_total_energy: float = None
     ccsd_prt_pr_dipole_moment: List[float] = None
 
-    # CCSDT keywords
-    ccsdt_correlation_energy: float = None
-    ccsdt_total_energy: float = None
-    ccsdt_dipole_moment: List[float] = None
-    ccsdt_iterations: int = None
-
-    # CCSDTQ keywords
-    ccsdtq_correlation_energy: float = None
-    ccsdtq_total_energy: float = None
-    ccsdtq_dipole_moment: List[float] = None
-    ccsdtq_iterations: int = None
-
     class Config:
         allow_mutation = False
         extra = "forbid"
--- a/qcelemental/models/results.py
+++ b/qcelemental/models/results.py
@@ -37,7 +37,7 @@ class ResultProperties(BaseModel):
     mp2_opposite_spin_correlation_energy: float = None
     mp2_singles_energy: float = None
     mp2_doubles_energy: float = None
-    mp2_total_correlation_energy: float = None
+    mp2_correlation_energy: float = None
     mp2_total_energy: float = None
     mp2_dipole_moment: List[float] = None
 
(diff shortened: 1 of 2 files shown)


dgasmith:
I think this should a PR to QCSchema first. A few of those quantities will need string explanations with them and there will be some feedback there I am guessing.

sjrl:
I have made a PR to QCSchema for the CCSD quantities (https://github.com/MolSSI/QCSchema/pull/61).

dgasmith:
This should be good to update.

sjrl:
So I have added the CC keywords to QCElemental. Should I also go ahead with the renaming of `mp2_total_correlation_energy` to `mp2_correlation_energy`?

loriab:
> Should I also go ahead with the renaming of mp2_total_correlation_energy to mp2_correlation_energy?

yes, please. any and all changes to make qcel agree with the qcschema PR.

sjrl:
Okay I've updated the mp2 to match the QCSchema PR. Presumably this will cause some problems in other parts of the QCArchive.

qcelemental/models/results.py
```diff
@@ -37,8 +37,36 @@ class ResultProperties(BaseModel):
...
+    # CCSD(T) keywords
+    ccsd_prt_pr_correlation_energy: float = None
+    ccsd_prt_pr_total_energy: float = None
+    ccsd_prt_pr_dipole_moment: List[float] = None
+
+    # CCSDT keywords
+    ccsdt_correlation_energy: float = None
+    ccsdt_total_energy: float = None
+    ccsdt_dipole_moment: List[float] = None
+    ccsdt_iterations: int = None
```
dgasmith: CCSDT(Q) last one (ugh).

sjrl: Is this for completions sake? Since it looks like there a number of other CC methods out there too.

dgasmith: Yea, honestly scratching my head on this one a bit after seeing it in practice. Anything beyond CCSD(T) might be better served in `qcvars`. @loriab?

loriab: just checked and CCSDT(Q) wasn't added to qcschema. I think it's ok to skip for now.

sjrl: Yeah it doesn't seem there are a lot of codes out there that support past CCSD(T). I guess if we are still on this topic I'd only advocate for the addition of CC-F12 methods.

loriab: in my experience, [these](https://github.com/qcdb/qcdb/blob/master/qcdb/psivarrosetta.py#L158-L183) are what you want to collect for f12 methods from molpro. not sure what balance of work-with-this-and-learn-from-practice or add-em-now to suggest.

loriab: oh, and one shouldn't actually be storing (T), (T*), (T**), as those can be built by equation

sjrl:
I'm happy to leave as is and add in more in later rounds. So to clarify do I need to add a line for the current PR to https://github.com/MolSSI/QCElemental/blob/master/docs/source/changelog.rst ?

dgasmith:
Lets leave it as is, but snip off everything after CCSD(T). The changeling line would be great!

sjrl:
Okay so snip out stuff after CCSD(T) here in `results.py` but leave those quantities in QCSchema? And great I'll add a line.

PR #56, opened 2019-04-16
B787 dreamaligner

loriab:
This is an initial drop-in of the alignment tools. Expect it'll fail CI b/c I didn't add in optional networkx or the like. coverage will probably plummet b/c so many helper fns. Don't worry about the B787 fn itself being so complicated to use -- it simplifies a lot when hooked up to a Mol-class-like like `models.Molecule`.

This is a good point to discuss deps and suitability for qcel.


dgasmith:
I think Networkx would be a bit too heavy as a required dependency here. Optional could be ok however.

loriab:
Definitely optional for networkx. Ok with the hefty number of her hyper fns? They’re all well written except for uno.

dgasmith:
Im fine with it as long as we can boost the coverage to something reasonable. Additional tested code without depends is fine here.

qcelemental/util/gph_uno_bipartite.py
```diff
@@ -0,0 +1,811 @@
...
+Author: guangzhi XU (xugzhi1987@gmail.com; guangzhi.xu@outlook.com)
+Update time: 2017-06-29 14:41:56.
+
+Updated Dec 2017 LAB for pep8, py3, more tests, starter_match, and simpler interface
+
+"""
+from __future__ import print_function
+
+try:
+    import networkx as nx
```
dgasmith: We will need to move this into the functions themselves. Also optional depends like this should be lazy loaded like `pint` is.

loriab: right now the whole `gph_uno_bipartite.py` is conditionally loaded only if one enters the alignment code (that is, uno not in `util/__init__.py`). Prefer it always loaded and `import networkx as nx` in half-a-dozen fns over current state?

dgasmith: I would always vote for importing the files so that structure is not lazy loaded. I agree its messier for programmers, but I think eliminates some lazy loading edge cases.

loriab:
This is about as good as I can get the coverage. It's ready for consideration.

Note that the usage case I heard (aligning 4 atoms of a molecule to the torsion scan 4 atoms) will need additional subset pre work b/c aligner presently expects two molecules (or sets of arrays) with same atom type composition.

qcelemental/models/molecule.py
```diff
@@ -723,3 +723,235 @@ def nelectrons(self, ifr: int = None) -> int:
             nel = sum([zf for iat, zf in enumerate(Zeff) if iat in self.fragments[ifr]]) - self.fragment_charges[ifr]
 
         return int(nel)
+
+    def B787(concern_mol,
+             ref_mol,
+             do_plot=False,
+             verbose=1,
```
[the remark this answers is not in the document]
loriab: Are you suggesting it should be quiet by default or that the meanings of the levels should change? I usually use 1 for ordinary printing and 0 for quiet/suppress.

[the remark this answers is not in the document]
dgasmith: Quiet by default.

qcelemental/models/molecule.py
```diff
@@ -723,3 +723,235 @@ def nelectrons(self, ifr: int = None) -> int:
             nel = sum([zf for iat, zf in enumerate(Zeff) if iat in self.fragments[ifr]]) - self.fragment_charges[ifr]
 
         return int(nel)
+
+    def B787(concern_mol,
```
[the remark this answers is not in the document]
loriab: * it's you and CDS vs. me and Carlos on `align` vs. `B787` :-)  But I yield to find-ability.
* sure on the `(mol, data)` return
* note that traditionally, it's `amol = cmol.align(rmol)`, where `C` is changeable/concern, `R` is reference/riveted and `A` is aligned/answer. very easy to get jumbled with `self`. so I'd like to stick with that, which is the opposite of your "copy of mol2/rmol aligned as best as possible to mol1/cmol"

qcelemental/molparse/align.py
```diff
@@ -0,0 +1,716 @@
...
+
+def _pseudo_nre(Zhash, geom):
+    """Pseudo nuclear repulsion energy where non-physical Z contrived from `Zhash`."""
+
+    Zidx = list(set(sorted(Zhash)))
+    pZ = [Zidx.index(z) for z in Zhash]
+    return _nre(pZ, geom)
+
+
+#         1         2         3         4         5         6         7         8         9         10
```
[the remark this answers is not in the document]
loriab: that's my docstring ruler. now gone.

loriab:
this'll need rebasing and adjusting after #58, but other items have been addressed.

loriab:
RTG after CI

dgasmith:
Can you patch up LGTM before merging?

qcelemental/models/molecule.py
```diff
@@ -723,3 +709,235 @@ def nelectrons(self, ifr: int = None) -> int:
...
+                quiet=(verbose > 1))
+            assert compare(
+                True,
+                np.allclose(ref_mol.geometry, amol.geometry, atol=4),
+                'Q: concern_mol-->returned_mol geometry matches ref_mol',
+                quiet=(verbose > 1))
+
+        return amol, {'rmsd': rmsd, 'mill': solution}
+
+    def scramble(ref_mol,
```
dgasmith: This is a pure test function and should go in tests. If it just scrambled then it would likely be fine.

loriab: How about a return bool so it could scramble or scramble-and-test? problem is that knowing what to test is best kept internal, but compared to the pure scramble fn in molparse/align, it's nice to have it hooked up to Mol.

dgasmith: It doesn't serve any useful purpose to a user as implemented?

loriab: correct. if the user wanted an instance scrambled, my proposed `def scramble(ref_mol, do_shift, ..., do_test=False) -> Molecule` would return a new instance, scrambled accordingly, or a `bool` if `do_test=True` (`return_bool=True`?)

dgasmith: But why have the test here in the first place? This could be a function in the test file rather than here. I don't see other instances of code which have tests within functions.

loriab:
> Can you patch up LGTM before merging?

I'd like to, only it's ignoring me for 2 of them, and I'm not following 2 of its in return.

loriab:
RTG after CI. If it doesn't cancel the lgtm alert this time, I don't know what to do -- it seems stuck.

Screen Shot 2019-04-25 at 4 03 33 PM
Screen Shot 2019-04-25 at 4 02 40 PM

loriab:
As I finally learn from big green box https://lgtm.com/help/lgtm/alert-suppression, lgtm suppression never takes effect on PRs, only upon merge. So RTG.

PR #58, opened 2019-04-23
importing utils

loriab:
migrate `which`, `which_import`, etc. from qcng for use in soft import of deps and testing

I tried putting them in `util/` but that caused import problems. didn't try too hard to detangle. figure `importing` can be like `testing`.


qcelemental/physical_constants/__init__.py
```diff
@@ -3,12 +3,10 @@
 """
 
 # Ensure pint exists
-from importlib.util import find_spec
-spec = find_spec('pint')
-if spec is None:
+from ..importing import which_import
+if not which_import('pint', return_bool=True):
```
dgasmith: Would rather leave these in `util` and not bother with the import here.

loriab: then not going to get the soft "pint not found. do `pip install pint`" msg?

dgasmith: Simply revert to the previous way of doing things?

loriab: revised

qcelemental/physical_constants/__init__.py
```diff
@@ -2,13 +2,4 @@
...
 """
 
-# Ensure pint exists
-from importlib.util import find_spec
-spec = find_spec('pint')
-if spec is None:
-    raise ModuleNotFoundError(
-        """Python module pint not found. Solve by installing it: `conda install pint -c conda-forge` or `pip install pint`"""
-    )
-del spec, find_spec
```
dgasmith: I think there is some misunderstanding. I don't think we should edit this file at all and leave as it was before this PR. This is the optimal solution and the small duplication of code isn't worth the additional delayed importing hassles.

qcelemental/util/importing.py
```diff
@@ -0,0 +1,67 @@
+import os
```
dgasmith: Great place for this!

loriab:
this should be RTG after CI

dgasmith:
Great! Thanks.

PR #59, opened 2019-04-25
Update results.py

dgasmith:
We need to stage out these changes rather than doing it in one pass. Would be *very* rough to do without a transition release.


loriab:
worth doing an `UpgradeHelper` https://github.com/psi4/psi4/pull/1351/files#diff-a7c88e038c7c4c8ab84033e68c2b7c38R99 https://github.com/psi4/psi4/blob/master/psi4/driver/p4util/exceptions.py#L85 or more targeted at database than users?

dgasmith:
It's a database issue.

PR #62, opened 2019-05-07
typing and star pass

loriab:
I set out to preventatively add the `*` in the kwargs list to force keyword-only args for easier future dev. And I thought I'd get a little typing under my fingers by filling in fn signatures. Then I realized I could be adding errors since those sigs are never checked (unless in pydantic class), so started `mypy` linting. I've fixed many alerts, but they're so numerous that I'd better get a gauge on what's wanted before proceeding.

### DIFF
--- a/qcelemental/molparse/nucleus.py
+++ b/qcelemental/molparse/nucleus.py
@@ -1,4 +1,5 @@
 import re
+from typing import List, Tuple
 
 from .regex import NUCLEUS
 from ..exceptions import NotAnElementError, ValidationError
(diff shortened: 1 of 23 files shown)


dgasmith:
I think this is all good things. Getting `mypy` working at some point would be great, but not in a huge hurry to do so at the moment.

loriab:
This is the end of my current typing push. There are some legitimate fixes in here, so good to get it in and stabilized a bit before the next release.

qcelemental/molparse/nucleus.py
```diff
@@ -248,17 +249,17 @@ def offer_user_label(lbl):
 
     text = ['', """--> Inp: A={}, Z={}, E={}, mass={}, real={}, label={}""".format(A, Z, E, mass, real, label)]
 
-    Z_exact = []  # *_exact are candidates for the final value
-    Z_range = []  # *_range are tests that the final value must pass to be valid
-    A_exact = []
-    A_range = []
-    m_exact = []
-    m_range = []
+    Z_exact: List = []  # *_exact are candidates for the final value
```
dgasmith: Any reason to type these inside functions?

loriab: only that mypy raised alerts and I knew how to fix them

dgasmith: Gotcha.

PR #63, opened 2019-05-07
importing raise_error and molutil

loriab:
- [x] I'm not wholly convinced, but add `raise_error` for default to `which` and `which_import` after discussion in https://github.com/MolSSI/QCEngine/pull/65
- [x] Replace deprecated `find_loader` with `find_spec`.
- [x] Q: `ModuleNotFoundError` is _not_ right for `which` but what is?
- [x] Move numpy alignment to molutil

### DIFF
--- a/qcelemental/tests/test_importing.py
+++ b/qcelemental/tests/test_importing.py
@@ -1,3 +1,5 @@
+import pytest
+
 import os
 
 import qcelemental as qcel
@@ -23,6 +25,13 @@ def test_which_import_f_bool():
     assert ans is False
 
 
+def test_which_import_f_raise():
+    with pytest.raises(ModuleNotFoundError) as e:
+        qcel.util.which_import('evilpint', raise_error=True)
+
+    assert "'evilpint' not found" in str(e)
+
+
 def test_which_t():
     ans = qcel.util.which('ls')
     assert ans.split(os.path.sep)[-1] == 'ls'
@@ -36,6 +45,13 @@ def test_which_t_bool():
 def test_which_f():
     ans = qcel.util.which('evills')
     assert ans is None
+
+
+def test_which_f_raise():
+    with pytest.raises(ModuleNotFoundError) as e:
+        qcel.util.which('evills', raise_error=True)
+
+    assert "'evills' not found" in str(e)
 
 
 def test_which_f_bool():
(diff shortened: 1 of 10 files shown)


qcelemental/tests/test_importing.py
```diff
@@ -23,6 +25,13 @@ def test_which_import_f_bool():
     assert ans is False
 
 
+def test_which_import_f_raise():
+    with pytest.raises(ModuleNotFoundError) as e:
+        qcel.util.which_import('evilpint', raise_error=True)
+
+    assert "'evilpint' not found" in str(e)
```
dgasmith: Hmm... no such thing.

qcelemental/util/importing.py
```diff
@@ -40,12 +48,20 @@ def which(command, *, return_bool=False):
...
+
     """
-    lenv = {'PATH': ':' + os.environ.get('PATH') + ":" + os.path.dirname(sys.executable)}
+    lenv = {'PATH': ':' + os.environ.get('PATH', '') + ':' + os.path.dirname(sys.executable)}
     lenv = {k: v for k, v in lenv.items() if v is not None}
 
     ans = shutil.which(command, mode=os.F_OK | os.X_OK, path=lenv['PATH'])
 
+    if raise_error and ans is None:
+        raise ModuleNotFoundError(f"Command '{command}' not found in envvar PATH.")
```
dgasmith: I think you convinced me of this one where we should always add custom messages. What if we added:

`"Command '{command}' not found in envvar PATH. {if install else ''}"`

or similar here? Would hide a bit more logic.

loriab: Where `install` is a message passed into the fn? I decided against that once, but not strenuously. Fine by me.

loriab:
this is RTG

dgasmith:
Squish LGTM commits?

loriab:
squish the whole thing, imo

loriab:
Fixed some `os.pathsep` and added an `env` to `which` to help code with custom envvar. Not sure I did it the best way. Here's a context manager: https://github.com/laurent-laporte-pro/stackoverflow-q2059482/blob/master/demo/environ_ctx.py

dgasmith:
LGTM

PR #64, opened 2019-05-08
Molecular Visualization

dgasmith:
Moves visualize to `show` command. Reason for this is I think the base should `show` so that it is easier to type out many different types of visualization for a single object while keeping `visualize` to bring up more of a dashboard like object portal side. Examples:
 - `show_trajectories/show_final_moleucle/show_changes` for optimizations.
 - `show_ternary/show_statistics/etc` on reactiondatasets.
 - `show_statistics/show_torsions/show_counts/etc` on torsion drive datasets.

For the reactiondatasets `visualize()` would bring up a widget of show statistics that would offer user clickable options rather than having to sift and write your own `method=[B3lyp,....]` lines. `visualize` would still of course use `show` commands under the hood.

Any thoughts on this name change?

Also adds better repr and automatic molecular visualization in notebooks.

### DIFF
--- a/qcelemental/models/molecule.py
+++ b/qcelemental/models/molecule.py
@@ -101,7 +101,7 @@ class Molecule(BaseModel):
 
     # Extra
     provenance: Provenance = provenance_stamp(__name__)
-    id: Optional[str] = None
+    id: Optional[Any] = None
     extras: Dict[str, Any] = None
 
     class Config:
@@ -658,8 +658,18 @@ class Molecule(BaseModel):
                 break
         return new_geometry
 
+
     def __str__(self):
         return self.pretty_print()
+
+    def __repr__(self):
+        return f"<{self.__class__.__name__}(name='{self.name}' formula='{self.get_molecular_formula()}')>"
+
+    def _repr_html_(self):
+        try:
+            return self.show()._repr_html_()
+        except ModuleNotFoundError:
+            return f"<p>{repr(self)[1:-1]}</p>"
 
     @staticmethod
     def _inertial_tensor(geom, *, weight):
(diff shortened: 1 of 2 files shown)


qcelemental/models/molecule.py
```diff
@@ -214,8 +214,8 @@ def json_dict(self, *args, **kwargs):
 
 ### Non-Pydantic API functions
 
-    def visualize(self, *, style: Union[str, Dict[str, Any]] = "ball_and_stick",
-                  canvas: Tuple[int, int] = (400, 400)) -> 'py3Dmol.view':
+    def show(self, style: Union[str, Dict[str, Any]] = "ball_and_stick",
+             canvas: Tuple[int, int] = (400, 400)) -> 'py3Dmol.view':
```
loriab: style/canvas as positional or keyword args?

dgasmith: Always keyword, good point.

Dashboard could be good as well over visualize as that is more of what it is moving too.

For `plot` would we want `show_reaction` and `plot_statistics` or both show?

loriab: I think choose one or the other btwn show & plot. Predictable API more valuable than nature of the viz. "show" is fine with me -- I just think it could be guessed as "print", whereas "plot" definitely suggests additional backend.

I like "dashboard" a little better than "visualize" b/c more concrete. But good with either.

loriab: > For plot would we want show_reaction and plot_statistics or both show?

Ok, I read this again and realized you were making a distinction btwn chemistry and chemistry+data. That weakens my desire for consistency, but I think I still favor both as show/plot/display.

dgasmith: I agree on `show` for both as well. Something that will be changed in other repositories, but worth thinking about here I think.

PR #68, opened 2019-05-28
add options info to molparse.to_string

loriab:
I've gone back and forth on this over the years -- whether to combine or separate molecule formatting. I suspect it'll help qcengine to combine, so proposing this to collect opinions, namely @dgasmith @sjrl @ffangliu.

I think the idea is pretty clear -- in addition to stuffing as much molecule info as available into the string (which isn't nearly all in most `dtype`s), additionally return the options that the molecule knows about, and, for the downstream programmer, a record of what fields have been expressed.

EDIT:
- [x] `Datum` gets a `numeric` field set by validation. If user wants non-numeric, have to pass `numeric=False` flag at initialization.


qcelemental/datum.py
```diff
@@ -26,18 +26,29 @@ class Datum(BaseModel):
...
     glossary: str = ''
 
-    def __init__(self, label, units, data, *, comment=None, doi=None, glossary=None):
-        kwargs = {'label': label, 'units': units, 'data': data}
+    class Config:
+        extra = "forbid"
+        allow_mutation = False
+        json_encoders = {
+            np.ndarray: lambda v: v.flatten().tolist(),
+            complex: lambda v: (v.real, v.imag),
```
dgasmith: Looks like no built in Pydantic support for complex numbers, we might consider adding it there.

loriab: that'd be great. especially as this is going to cause trouble for reconstituting (complex or tuple?).

I'd like to get this is as a stopgap for now so I can clear out some PRs.

PR #70, opened 2019-06-03
Molpro: Added dtype molpro to molecule.to_string function.

sjrl:
Added molpro dtype to molecule.to_string so I can update the `MolproHarness` `build_input`

### DIFF
--- a/qcelemental/molparse/to_string.py
+++ b/qcelemental/molparse/to_string.py
@@ -21,7 +21,7 @@ def to_string(molrec: Dict,
     ----------
     molrec : dict
         Psi4 json Molecule spec.
-    dtype : {'xyz', 'cfour', 'nwchem'}
+    dtype : {'xyz', 'cfour', 'nwchem', 'molpro'}
         Overall string format. Note that it's possible to request variations
         that don't fit the dtype spec so may not be re-readable (e.g., ghost
         and mass in nucleus label with ``'xyz'``).
@@ -71,6 +71,7 @@ def to_string(molrec: Dict,
         "xyz": "Angstrom",
         "cfour": "Bohr",
         "gamess": "Bohr",
+        "molpro": "Bohr",
         "nwchem": "Bohr",
         "psi4": "Bohr",
         "terachem": "Bohr"
@@ -142,6 +143,31 @@ def to_string(molrec: Dict,
             'units': umap.get(units.lower()),
             'coordinates': 'cartesian',
         }
+
+    elif dtype == 'molpro':
+
+        atom_format = '{elem}'
+        ghost_format = '{elem}'
+        umap = {'bohr': 'bohr', 'angstrom': 'angstrom'}
+
+        atoms = _atoms_formatter(molrec, geom, atom_format, ghost_format, width, prec, 2)
+
+        ghost_atoms = []
+        nat = geom.shape[0]
+        for iat in range(nat):
+            if not molrec['real'][iat]:
+                ghost_atoms.append(str(iat+1))
+
+        first_line = f"""{{{umap.get(units.lower())}}}"""
+        second_line = """geometry={"""
+        end_bracket = """}"""
+        smol = [first_line, second_line]
+        smol.extend(atoms)
+        smol.append(end_bracket)
+
+        if len(ghost_atoms) != 0:
+            ghost_line = f"""dummy,{",".join(ghost_atoms)}"""
+            smol.append(ghost_line)
 
     elif dtype == 'nwchem':
 
--- a/qcelemental/molparse/to_string.py
+++ b/qcelemental/molparse/to_string.py
@@ -152,12 +152,6 @@ def to_string(molrec: Dict,
 
         atoms = _atoms_formatter(molrec, geom, atom_format, ghost_format, width, prec, 2)
 
-        ghost_atoms = []
-        nat = geom.shape[0]
-        for iat in range(nat):
-            if not molrec['real'][iat]:
-                ghost_atoms.append(str(iat+1))
-
         first_line = f"""{{{umap.get(units.lower())}}}"""
         second_line = """geometry={"""
         end_bracket = """}"""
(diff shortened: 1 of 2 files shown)


qcelemental/molparse/to_string.py
```diff
@@ -143,6 +144,21 @@ def to_string(molrec: Dict,
             'coordinates': 'cartesian',
         }
 
+    elif dtype == 'molpro':
+
+        atom_format = '{elem}'
+        ghost_format = 'Q'
```
loriab: sure that's not a dummy? I found different instructions for ghost in my old notes (below). May as well construct the charge, mult, dummy commands here, I think.

```

# setup some QC program differences
if ( ($method eq "xyz") || ($method eq "qchem") || ($method eq "xdm") ) {

   $gh  = "Gh";   # to ghost an atom, replace its symbol with this and specify the atom type in the basis list
   $dm  = "X ";   # dummy atoms stored this way in resource file zmat
   $eql = "=";    # separate variable name and value in zmats
}
elsif ($method eq "molpro") {

   $gh  = "NaN";  # to ghost an atom, leave it alone in the geometry spec but feed its number to the dummy card
   $dm  = "Q ";   # can be Q or X, but the former can't be confused with symmetry specs
   $eql = "=";
   $memory = $memory/8.0;  # translate megabytes to megawords for molpro
}
elsif ( ($method eq "psi4") || ($method eq "sapt") || ($method eq "database") ) {

   $gh  = "NaN";  # to ghost an atom, leave it alone in the geometry and pick out the submolecules
   $dm  = "X ";   # looks to be X from the sample files
   $eql = "=";
}
elsif ($method eq "nwchem") {

   $gh  = "Bq";   # to ghost an atom, start it with Bq and specify a basis set for example, BqH or BqO
   $dm  = "X ";   # can be Bq or X, musn't have basis functions attached HOWEVER, doesn't actually like Bq with zmats
   $eql = " ";
}
elsif ($method eq "gaussian") {

   $gh  = "Bq";
   $dm  = "X ";
   $eql = " ";
}
```

qcelemental/molparse/to_string.py
```diff
@@ -143,6 +144,21 @@ def to_string(molrec: Dict,
...
+        umap = {'bohr': 'bohr', 'angstrom': 'angstrom'}
+
+        atoms = _atoms_formatter(molrec, geom, atom_format, ghost_format, width, prec, 2)
+
+        first_line = """{{{}}}""".format(umap.get(units.lower()))
```
loriab: feel free to use f-strings, as qcel is py36+

qcelemental/molparse/to_string.py
sjrl: Hmm I suppose I need clarification on dummy atoms vs ghost atoms. It appears if you use Q or X in a Z matrix format it makes a center with no charge or basis functions (https://www.molpro.net/develop/master/doc/manual/node103.html). Otherwise if you want the atom to have basis functions but no charge you need to call the dummy card as you mention in your notes. 

So which should the ghost atom in QCElemental case refer to?

loriab: The broadly used definitions are as below. molpro's "dummy" list is badly named.
* atom -- yes charge, yes basis fns
* ghost -- no charge, yes basis fns
* dummy -- is a point, no charge, no basis fns

qcschema.molecule doesn't acknowledge zmat, so everything's in cartesian and atom & ghost are all we're concerned with.

sjrl: Thanks for the clarification! I'll switch over `ghost_format = '{elem}'` and then add the appropriate dummy card for ghost atoms.

sjrl: Is there an easy way to accumulate the specified ghost atoms, so I can them write them to the dummy card?

loriab: Molrec[‘real has the info you want. Like models.molecule.real, it’s a list of bools

dgasmith:
@loriab @sjrl Anything else that needs to be done here?

sjrl:
Yes I still need to update the ghost atom specification for Molpro. Currently working on that now.

sjrl:
I've finished adding the ghost atom specification in Molpro.

qcelemental/molparse/to_string.py
```diff
@@ -143,6 +144,31 @@ def to_string(molrec: Dict,
...
+            if not molrec['real'][iat]:
+                ghost_atoms.append(str(iat+1))
+
+        first_line = f"""{{{umap.get(units.lower())}}}"""
+        second_line = """geometry={"""
+        end_bracket = """}"""
+        smol = [first_line, second_line]
+        smol.extend(atoms)
+        smol.append(end_bracket)
+
```
loriab: could you add the `set charge` and `mult` commands in here, too?

I'm favoring `to_string` returning all text that can be placed adjacent to the Cartesian lines, so I think chg/mult/dummy is in that.

qcelemental/molparse/to_string.py
```diff
@@ -143,6 +144,31 @@ def to_string(molrec: Dict,
...
+        if len(ghost_atoms) != 0:
+            ghost_line = f"""dummy,{",".join(ghost_atoms)}"""
+            smol.append(ghost_line)
```
loriab: can be a little more compact

```
reals = [True, False, True, True, False]  # molrec['real']

ans = 'dummy,' + ','.join([str(idx + 1) for idx, real in enumerate(reals) if not real])

print(ans)
# dummy,2,5
```

qcelemental/molparse/to_string.py
sjrl: Yup that sounds good to me. The only thing that could potentially be problematic is we ever support specifying a separate xyz file to be read. But that's probably not worth worrying about right now.

qcelemental/molparse/to_string.py
sjrl: I've updated the code to make it more compact

qcelemental/molparse/to_string.py
```diff
@@ -143,6 +144,28 @@ def to_string(molrec: Dict,
...
+        end_bracket = """}"""
+        smol = [first_line, second_line]
+        smol.extend(atoms)
+        smol.append(end_bracket)
+
+        if False in molrec['real']:
+            ghost_line = 'dummy,' + ','.join([str(idx + 1) for idx, real in enumerate(molrec['real']) if not real])
+            smol.append(ghost_line)
+
+        smol.append(f"set,charge={molrec['molecular_charge']}")
```
loriab: if Molpro only accepts integer charges, will need to `int(...)` this.

qcelemental/molparse/to_string.py
loriab: Agreed, I think xyz file is template only.

qcelemental/molparse/to_string.py
sjrl: It seems like Molpro reads charge as a float, but I'm not sure if it still expects it to be an integer charge.

loriab: if it reads as a float, then leave it. running or not is their responsibility then.

sjrl:
Okay tests have been added.

PR #74, opened 2019-06-14
preserve atom and fragment order with get_fragment()

### LINKED ISSUE #75: orientation and ordering in get_fragment
The [`get_fragment()`](https://github.com/MolSSI/QCElemental/blob/master/qcelemental/models/molecule.py#L336-L412) fn has been working nicely for fractal fragment filtering for a while. It has two issues:

* (a) the returned mol always has real frags first, followed by ghosted frags. so if enough things match, `get_fragment(1, 0)` and be the same mol (down to atom ordering) as `get_fragment(0, 1)`. this will surprise users of `psi4.Molecule.extract_subsets()` who expect fragment ordering to be preserved. The latter behavior is being implemented in #74.

* (b) the current fn comes with an `orient=False` default. But that's not getting passed through to anywhere, so probably [default False](https://github.com/MolSSI/QCElemental/blob/master/qcelemental/models/molecule.py#L112) is imposed. Easy to pass this flag along but probably should be careful of downstream.

loriab:
closes #75 

see there for description of problem.

### DIFF
--- a/qcelemental/models/molecule.py
+++ b/qcelemental/models/molecule.py
@@ -333,11 +333,35 @@ class Molecule(BaseModel):
 
         return text
 
-    def get_fragment(self, real: Union[int, List], ghost=None, orient: bool = False):
-        """
-        A list of real and ghost fragments:
-        """
+    def get_fragment(self,
+                     real: Union[int, List],
+                     ghost: Optional[Union[int, List]] = None,
+                     orient: bool = False,
+                     group_fragments: bool = True) -> 'Molecule':
+        """Get new Molecule with fragments of self real, ghosted, or dropped.
 
+        Parameters
+        ----------
+        real
+            Fragment index or list of indices (0-indexed) to be real atoms in new Molecule.
+        ghost
+            Fragment index or list of indices (0-indexed) to be ghost atoms (basis fns only) in new Molecule.
+        orient
+            Whether or not to align (inertial frame) and phase geometry upon new Molecule instantiation
+            (according to _orient_molecule_internal)?
+        group_fragments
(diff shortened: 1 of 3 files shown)


loriab:
preliminary, as I want to add a test with a skipped frag to be sure operational.

this is a breaking change.

if it's going to interfere with fractal, we'd better put in a flag that chooses btwn the two schemes. the current scheme is going to surprise psi4 `extract_subset` users.

dgasmith:
We will need a flag here to switch between the two schemes. Maybe `orient=True, reorder=True`?

loriab:
maybe `contiguize_and_reorder=True` to be obvious? otherwise, I'm fine with `reorder` so long as the docstring is very clear. Even the contiguize... doesn't mention the real/ghost grouping.

dgasmith:
Separate options would be best I believe? Yes, the doc strings are lacking and should be added.

qcelemental/models/molecule.py
```diff
@@ -333,11 +333,35 @@ def pretty_print(self):
...
 
-    def get_fragment(self, real, ghost=None, orient=False):
-        """
-        A list of real and ghost fragments:
-        """
+    def get_fragment(self,
+                     real: Union[int, List],
+                     ghost: Optional[Union[int, List]] = None,
+                     orient: bool = False,
+                     group_fragments: bool = True) -> 'Molecule':
```
dgasmith: What do you think about changing the default behavior of this to False after we update Fractal?

loriab: Yes, I'm all for that so long as Fractal stays happy. The current default is, to me, unexpected and slippery for gradients.

PR #78, opened 2019-06-26
Vector gradient

psi-rking:
I added function to align the derivatives of vector components.  For example, the code
will align dipole moment derivatives.  It can convert CFOUR's DIPDER with PSI4's file17 values.

A remaining danger is that CFOUR reorders atoms when computing the gradients.  Unless one attends carefully to the atomic numbers in DIPDER, or processes carefully the symmetry sorting presented in the output file, one might not even notice.  I've attached a sample output and dipder file.  The geometry does not appear to be printed anywhere in the order that was used in DIPDER, which means reliable atom mapping will be a challenge.

[c4-vib.txt](https://github.com/MolSSI/QCElemental/files/3331894/c4-vib.txt)
[c4-dipder.txt](https://github.com/MolSSI/QCElemental/files/3331895/c4-dipder.txt)

The dipole moment derivatives are presented in the output file too.  Only to 6 decimals.

Does the same problem emerge in a Hessian computation?  Perhaps you have solved it.


qcelemental/molutil/test_align.py
```diff
@@ -442,3 +442,77 @@ def test_hessian_align():
 
     p2chess = mill.align_hessian(p4_hooh_hess_native)
     assert compare_values(c4_hooh_hess, p2chess, atol=1.e-4)
+
+def test_vector_gradient_align():
```
loriab: because `atoms_map=False` invokes the reordering machinery, need `@using_networkx` for this test like [here](https://github.com/MolSSI/QCElemental/pull/78/files#diff-c80b22cde9934b3725bea9e0aa7dcf1cR206).

qcelemental/models/align.py
```diff
@@ -126,6 +126,27 @@ def align_hessian(self, hess) -> NDArray:
...
+        for at in range(nat):
+            Datom.fill(0)
+            Datom[0,:] = mu_x[at,:]
+            Datom[1,:] = mu_y[at,:]
+            Datom[2,:] = mu_z[at,:]
+            Datom[:] = np.dot( self.rotation.T , np.dot(Datom, self.rotation) )
+            al_mu_x[self.atommap[at]][:] = Datom[0,:]
+            al_mu_y[self.atommap[at]][:] = Datom[1,:]
+            al_mu_z[self.atommap[at]][:] = Datom[2,:]
+        return (al_mu_x, al_mu_y, al_mu_z)
```
loriab: since inputs and outputs are xyz components, would it make sense to either do three separate calls or pass as [`(3 * nat, 3)`](https://github.com/psi4/psi4/blob/41f824cee253fc4262c08d7d0ae22190a889eae0/psi4/src/psi4/libmints/wavefunction.h#L245)? like to take better advantage of broadcasting.

qcelemental/models/align.py
```diff
@@ -126,6 +126,27 @@ def align_hessian(self, hess) -> NDArray:
...
+        al_mu_z = np.zeros( (nat, 3) )
+
+        Datom = np.zeros( (3,3) ) # atom whose nuclear derivatives are taken
+        for at in range(nat):
+            Datom.fill(0)
+            Datom[0,:] = mu_x[at,:]
+            Datom[1,:] = mu_y[at,:]
+            Datom[2,:] = mu_z[at,:]
+            Datom[:] = np.dot( self.rotation.T , np.dot(Datom, self.rotation) )
+            al_mu_x[self.atommap[at]][:] = Datom[0,:]
```
loriab: does the Cfour atom ordering for `DIPDER` match the geometry and gradient in `GRD`? I earnestly hope so, as that's where `atommap` gets computed from.

psi-rking: OK, yes they are the same.  Turns out that I was doing a bunch of second derivative computations and hadn't even noticed a GRD file being produced (in my scratch space).  Thus my concern.  A bit disturbing that the orientation can't be easily found in the output file, but it looks perhaps that CFOUR spits out the geometry to GRD even if it is not a gradient computation?  I will test the atom mapping now against the GRD coordinates and make sure it works.  Tricky for a user though who naively (naturally!) might think to provide the reoriented (but not reordered) geometry at the beginning of the output file.  Still, a step in the right direction.

loriab: if all else fails, I think the atom mapping is in `JOBARC` (which we can read in binary). but I'd prefer to stick with the find-the-mapping-and-orientation tech for consistency among programs.

psi-rking: I just figured out that the 'atommap' goes reverse mapping from target to original in direction from the mill.  I've got the atom sorting working now.

qcelemental/models/align.py
psi-rking: What I like about the 3 explicit arguments is that it is obvious to any user what should be passed, particularly for a mixed-derivative like this - versus the challenge of a new person (or myself in 6 months:) trying to figure out what the single variable data structure should be.  For the Hessian you would guess correctly, maybe not for this.  But I'm fine changing it if you think there's a compelling reason and relying on examples, tests, and documentation.

loriab: sorry for the delay -- it's been a crazy month. talking it over, we think the packed arrangement would be more in keeping with the other functions. the first line of the fn can be `mu_x, mu_y, mu_z = *mu` and the return can be `return al_mu` and the test call can be `p2c_dipder_x, p2c_dipder_y, p2c_dipder_z = mill.align_vector_gradient(...)`

qcelemental/molutil/__init__.py
```diff
@@ -1 +1,3 @@
 from .align import B787, kabsch_align, compute_scramble
+from .test_align import test_hessian_align, test_vector_gradient_align
```
loriab: this shouldn't be needed as we don't want the tests to be part of the module proper. if you let me know the error message that led to this addition, I can probably suggest an alternate solution.

psi-rking: I just wanted to be able to run the tests from an external directory.  If they are not here, how do you do that?   I was just testing with:
> sys.path.insert(0, '/home/rking/programs/QCElemental')
> import qcelemental as qcel
>qcel.molutil.test_vector_gradient_align()

loriab: Try just “pytest” once “import qcelemental as qcel” works.

PR #80, opened 2019-07-16
Update Change Log for 0.5.0

Lnaden:
Also bumps min Pydantic version to avoid bug from pydantic 0.30

### DIFF
--- a/devtools/conda-envs/base.yaml
+++ b/devtools/conda-envs/base.yaml
@@ -8,14 +8,14 @@ dependencies:
   - nomkl
   - python
   - pint
-  - pydantic>=0.20
+  - pydantic>=0.30.1
 
     # Optional depends
   - networkx
   - py3dmol
 
     # Testing
-  - pytest>=3.9.1
+  - pytest>=4.0.0
   - pytest-cov
   - codecov
   - scipy  # tests an aspect of a helper fn not used by qcel functionality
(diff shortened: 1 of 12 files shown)


Lnaden:
Is this the pytest 5+ changes to the error handling?

dgasmith:
Ugh yea.. should be `e.value`.

Lnaden:
This should fix the Pytest issues, but I think there is still an actual issue in `test_pubchem_multiout_g` throwing an error regarding the Choices that PubChem is returning (I think)

Lnaden:
I also verified that this works for Pytest 4.0+ and does not break anything downstream (i.e. Fractal)

dgasmith:
Looks like pubchem may have changed their interface slightly.

dgasmith:
I think I patched this.

Lnaden:
Probably, from what I can tell, the molecule we were using just isn't coming up in the search. There is also a mismatch between results from the direct search of the word `tropolone` on the website vs what comes up when we use the URL in the `pubchem.py` file. Might be a fluke?

dgasmith:
I think they just changed their database. It happens from time to time.

loriab:
Finding good example for pubchem is tricky and can change over time. Don’t feel need to preserve too strongly.

Lnaden:
Good to know, on both fronts. Thanks.

PR #81, opened 2019-07-17
Update base.yaml

loriab:
trivial, but I think we want to keep these fixed on oldest so it can signal when mins need to change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ install:
     # Create test environment for package
   - |
     if [ $PROG == "MINIMAL" ]; then
-      python devtools/scripts/conda_env.py -n=test -p=$PYTHON_VER devtools/conda-envs/minimal.yaml
+      python devtools/scripts/conda_env.py -n=test -p=$PYTHON_VER devtools/conda-envs/minimal_pins.yaml
     elif [ $PROG == "FULL" ]; then
       python devtools/scripts/conda_env.py -n=test -p=$PYTHON_VER devtools/conda-envs/base.yaml
     else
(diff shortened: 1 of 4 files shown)


Lnaden:
I would disagree, having a minimum lets us detect when upstream changes break things on our end through CI, and then we can fix it as it happens. If we pin to a specific version, it strongly limits the environment the package can be installed in with no real reason to enforce such a restriction.

loriab:
yikes, I meant to be editing [minimal](https://github.com/MolSSI/QCElemental/blob/master/devtools/conda-envs/minimal.yaml#L11), not base. do you agree with my intro comment applied to that file?

Lnaden:
I see. Not something I have ever done, but I can see how using the `minimal` file as a type of canary for when min versions break could be helpful. I think I would be okay with that on the `minimal` file, but I was confused at first as what the purpose of that environment was. 

Could we both update the README in that directory, add a small commented description in the file itself, and maybe rename it to `minumum_pins` or something? I see the merit of it now, but its not something I have personally encountered before so I want to make it abundantly clear what its function is for others in the future. Feel free to strike down any of those suggestions though as all of them may be overkill.

But you are still okay with the `base.yaml` file still having just lower bounds on pins?

loriab:
> But you are still okay with the base.yaml file still having just lower bounds on pins?

Definitely. It was entirely a mistake on my part to touch `base.yaml`. In fact, me not distinguishing between the base/minimal names is probably a signal to rename the latter.

> I see. Not something I have ever done, but I can see how using the minimal file as a type of canary for when min versions break could be helpful. I think I would be okay with that on the minimal file, but I was confused at first as what the purpose of that environment was.

Yes, this file first came about so that we'd notice if qcel didn't work when an optional dependency was missing. The pinning to minimal versions is an extra check.

I'll fix this up.

PR #84, opened 2019-07-21
don't test pubchem if no internet connection

loriab:
Thanks for the internet? snippet, @dgasmith. this now passes for me w/o wifi.


dgasmith:
Not sure what is going on with codecov.

loriab:
Yeah, was going to ask if you knew what was up with codecov. They don’t show outages. Will look into it.

PR #85, opened 2019-07-22
[WIP] ProtoModel base + Msgpack capabilities

dgasmith:
Begins a ProtoModel base class which all QCArchive models should inherit from.

A few notes:
 - Moves new Pydantic types new to `models/types.py` folder.
 - Adds msgpack serialization protocols with a custom NumPy array interface descriptor.
 - Moves most `List[x]` quantities to Array quantities.


qcelemental/util/serialization.py
```diff
@@ -0,0 +1,107 @@
...
+from .importing import which_import
+
+try:
+    import msgpack
+
+    _has_msgpack = True
+except ModuleNotFoundError:
+    msgpack = None
+
+    _has_msgpack = True
```
loriab: `False`? and is this var used?

qcelemental/models/molecule.py
```diff
@@ -91,17 +89,17 @@ class Molecule(BaseModel):
     molecular_multiplicity: int = 1
 
     # Atom data
-    masses: List[float]
-    real: List[bool]
-    atom_labels: Optional[List[str]] = None
-    atomic_numbers: Optional[List[int]] = None
-    mass_numbers: Optional[List[int]] = None
+    masses: Optional[Array[float]] = None
```
Lnaden: Small thing that is more a semantic thing. `Optional` and `= None` do the same thing from Pydantic's versions. But I wonder if having both makes it easier to understand from a user perspective. More of a thought than a request for change.

qcelemental/models/molecule.py
```diff
@@ -67,21 +69,17 @@ class Identifiers(BaseModel):
...
-        extra = "forbid"
-
-    def dict(self, *args, **kwargs):
-        return super().dict(*args, **{**kwargs, **{"skip_defaults": True}})
+    class Config(ProtoModel.Config):
+        serialize_skip_defaults = True
 
 
-class Molecule(BaseModel):
+class Molecule(ProtoModel):
```
Lnaden: So this change will make previous versions of Molecule incompatible due to the changes to List->Array (I think). Just a note for the change log more than anything.

qcelemental/models/results.py
```diff
@@ -83,21 +81,15 @@ class ResultInput(BaseModel):
...
 
-    class Config:
-        allow_mutation = False
-        extra = "forbid"
-
-        json_encoders = {**ndarray_encoder}
-
-    def json_dict(self, *args, **kwargs):
-        return json.loads(self.json(*args, **kwargs))
+    class Config(ProtoModel.Config):
```
Lnaden: I don't think we need this anymore.

qcelemental/models/types.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class ArrayMeta(type):
+    def __getitem__(self, t):
+        return type('Array', (TypedArray, ), {'_dtype': t})
+
+
+class Array(np.ndarray, metaclass=ArrayMeta):
+    def __init__(self, **kwargs):
+        super().__init__()
```
Lnaden: I might be missing something here, but is there reason we dont pass in the `**kwargs`? Similar, since this is a `np.ndarray`, are there not `*args` we need to capture and send as well?

Is the metaclass `ArrayMeta` taking care of that?

qcelemental/tests/test_molecule.py
```diff
@@ -240,10 +244,10 @@ def test_water_orient():
 
 
 def test_molecule_errors_extra():
-    data = water_dimer_minima.dict()
+    data = water_dimer_minima.dict(skip_defaults=True)
     data["whatever"] = 5
     with pytest.raises(Exception):
-        Molecule(**data)
+        Molecule(**data, validate=False)
```
Lnaden: Why do we validate=False here if we expect this to raise?

qcelemental/util/__init__.py
```diff
@@ -8,3 +8,5 @@
 from .internal import provenance_stamp
 from .itertools import unique_everseen
 from .importing import parse_version, safe_version, which, which_import
+
+from .serialization import msgpack_dumps, msgpack_loads
```
Lnaden: newline at end of file

qcelemental/models/basemodels.py
```diff
@@ -0,0 +1,120 @@
...
+        Returns
+        -------
+        Model
+            The requested model from a serialized format.
+        """
+
+        if content_type is None:
+            if isinstance(data, str):
+                content_type = "json"
+            elif isinstance(data, bytes):
```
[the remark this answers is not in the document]
dgasmith: It's a guess and the only way to guess. Supply the content_type if you are going to do something weird.

qcelemental/models/common_models.py
```diff
@@ -79,17 +53,15 @@ class Config:
...
     input_data: Any = None
     success: bool = False
     error: ComputeError
     extras: Optional[Dict[str, Any]] = None
 
-    class Config:
-        extra = "forbid"
-        allow_mutation = False
-        json_encoders = {**ndarray_encoder}
+    class Config(ProtoModel.Config):
```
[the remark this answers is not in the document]
dgasmith: Apparently this inheritance was fixed which is pretty nice!

qcelemental/models/common_models.py
```diff
@@ -2,40 +2,14 @@
...
-
-    @classmethod
-    def validate(cls, v):
-        try:
-            v = np.array(v, dtype=np.int)
-        except:
-            raise RuntimeError("Could not cast {} to NumPy Int Array!".format(v))
-        return v
+# Encoders, to be deprecated
+ndarray_encoder = {np.ndarray: lambda v: v.flatten().tolist()}
```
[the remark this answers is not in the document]
dgasmith: We cannot move this at the moment because downstream depends on it. As noted we want to patch that and remove it.

qcelemental/models/molecule.py
dgasmith: Does `= None` make it an optional? I recall trying this and running into issues.

qcelemental/models/molecule.py
dgasmith: Previous versions will upcast seamlessly because of the `np.asarray`.

qcelemental/tests/test_molecule.py
dgasmith: Because validate will clean the input arrays.

qcelemental/models/molecule.py
dgasmith: Right, so this is required for these cases. If you do `= None` it will assign the value to None during __init__; however, if you have `Optional` it will allow the value to be set to None in the after an init and by passing in `masses=None` to the constructor. This is a delicate issue, but required.

PR #86, opened 2019-07-27
Molecule validation performance improvements

dgasmith:
Previously a 300 atom, 300 step Optimization result would take ~10 seconds to validate and construct. This is now down to ~250ms with this PR. Major performance improvements:
 - Provides a LRU Cache for the nuclei validation (~700k function calls on 4 uniques).
 - Prevents string construction when the text is not used.
 - Faster distance detection by ~4-20x, and removes the check in most circumstances.
 - Prevents extraneous array copying.
 - Adds a validated flag to prevent duplicate validation, total construction time of the above mentioned optimization is now down to ~25ms.

This change will break QCFractal, some pinning ideas are required.


qcelemental/molparse/chgmult.py
```diff
@@ -490,28 +498,29 @@ def stringify(start, final):
...
                 ' and '.join(been_defaulted)))
 
-    if m_final != _high_spin_sum(fm_final):
+    if (m_final != _high_spin_sum(fm_final)) and log_brief:
         brief.append(
             '    Warning: Total multiplicity is not high-spin sum of fragments; may be clobbered by psi4.core.Molecule.update_geometry().'
         )
 
-    if verbose >= 2:
+    if log_full:
```
loriab: yikes, was forming those strings and appending them to `brief` or `text` slowing things down?

qcelemental/molparse/from_arrays.py
```diff
@@ -607,32 +615,34 @@ def validate_and_fill_nuclei(
         elea = np.asarray([None] * nat)
     else:
         # -1 equivalent to None
-        elea = np.array([(None if at == -1 else at) for at in elea])
+        elea = np.asarray(elea)
+        if None in elea:
```
loriab: should be `if -1 in elea:`?

qcelemental/molparse/to_schema.py
```diff
@@ -36,11 +36,14 @@ def to_schema(molrec: Dict, dtype: str, units: str='Bohr', *, np_out: int=False)
     """
     qcschema: Dict = {}
 
-    if molrec['units'] == 'Angstrom' and units == 'Bohr' and 'input_units_to_au' in molrec:
-        factor = molrec['input_units_to_au']
+    if units is None:
+        geom = np.array(molrec["geom"], copy=copy)
```
loriab: `factor` not getting defined in this block, which seems bad.

qcelemental/molparse/chgmult.py
dgasmith: Yea, about 30% of the total time or ~3 seconds in the above example.

qcelemental/molparse/to_schema.py
dgasmith: It is not being used anywhere else?

loriab: I think there'll be trouble in the case that you pass in `units=None, dtype=2` but molrec for this case happens to be Angstrom. As currently coded, you'll get an Ang `models.Molecule` 😱

dgasmith: Ah, ok. Let me try another way. Are we being too aggressive/dangerous in other optimization areas?

loriab: you mean in adding a `copy=False` option?

dgasmith: Other areas where we prevent copying as well.

qcelemental/models/molecule.py
```diff
@@ -131,7 +139,8 @@ def __init__(self, orient=False, validate=True, **kwargs):
         values = self.__values__
 
         natoms = values["geometry"].shape[0]
-        values["symbols"] = np.core.defchararray.title(self.symbols)  # Title case for consistency
+        if validate:
+            values["symbols"] = np.core.defchararray.title(self.symbols)  # Title case for consistency
```
loriab: are the "symbols" and "geometry" re-definitions in order to satisfy the skip_defaults note above? I was going to suggest that `if kwargs["validated"]:` would make clearer reading than `if validate:` but I guess they're functionally equivalent.

So if one deliberately declines validation, `GEOMETRY_NOISE` and `defchararray` don't get applied? I can see the former, as decimal stripping is "chemical", but code relies on the titlecase of the symbols -- ok to skip that?

qcelemental/molparse/from_schema.py
```diff
@@ -143,7 +154,7 @@ def contiguize_from_fragment_pattern(frag_pattern, *, geom=None, verbose:int=1,
             """Error: QCElemental would need to reorder atoms to accommodate non-contiguous fragments""")
 
     if geom is not None:
-        ncgeom = np.array(geom).reshape(-1, 3)
+        ncgeom = np.asarray(geom).reshape(-1, 3)
```
loriab: I suppose I expect a constructor to return an independent-memory object. Can't cite any particular danger, though. (note to self: l157 and l138)

qcelemental/models/molecule.py
dgasmith: `validated` is for shipping around Molecules that are already guaranteed to be passed through a Molecule class already. Ideally a user would never set `validated` themselves (or hopefully know about it).

As a note that `validate` and `validated` are different. `validate` can be overridden during init to force a validation on the input regardless of the `validated` kwarg.

dgasmith:
I like the idea of passing around a `copy=True/False` kwarg for all to/from functions so that there can be greater control over the data. In practice, I haven't seen this intermediate as much of an issue as Molecules are immutable and arrays are not mutated in-place in the code base. I do want to make sure the copies are minimized in some fashion as the previous version arbitrarily copied too much.

qcelemental/molparse/from_schema.py
dgasmith: See below, we can figure that one out.

loriab: sounds good on below. I like having a `copy=True/False` on all to/from rather than trying to foresee usage patterns. I was wishing for `const` for a bit. Freely admit that I err on the don't-surprise-me side of mem alloc, but now seems a good time to tighten up.

dgasmith: After hunting through this we added copy pretty much everywhere it is useful. I think I found the last ones... let me know.

Going to let the current `__init__` copy once.

qcelemental/models/basemodels.py
```diff
@@ -118,3 +118,6 @@ def msgpack(self,
     @classmethod
     def _parse_msgpack(cls, data: bytes) -> Dict[str, Any]:
         return msgpack_loads(data)
+
+    def to_string(self):
```
Lnaden: This overloads the built in `to_string` method of BaseModel. I'm okay with that, but it does have a different call signature, which may cause some confusion if someone tries to use it

qcelemental/physical_constants/context.py
```diff
@@ -178,6 +179,7 @@ def Quantity(self, data: str) -> 'Quantity':
 
         return self.ureg.Quantity(data)
 
+    @lru_cache()
```
Lnaden: Do you want to set an LRU cache `maxsize` here like was done with the reconcile_nucleus function?

dgasmith: I guessed the 128 was fine. Total number of conversion pairs used is small.

qcelemental/models/basemodels.py
dgasmith: I think its something that we can add back. The short is that the strings were not very useful  in our context so we needed to stop them as they cost a prodigious amount co compute. We often overload this individually in inherited projects.

dgasmith:
@loriab I think this is ready for final review.

qcelemental/molparse/to_schema.py
```diff
@@ -50,28 +58,29 @@ def to_schema(molrec: Dict, dtype: str, units: str='Bohr', *, np_out: int=False)
...
                 dtype, units))
-        qcschema = copy.deepcopy(molrec)
+        qcschema = deepcopy(molrec)
         qcschema['geom'] = geom
         qcschema['units'] = units
         qcschema['name'] = name
 
     elif dtype in [1, 2]:
-        if units != 'Bohr':
+        if units not in ['Bohr', None]:
```
loriab: `None` isn't an option, I don't think

qcelemental/molparse/from_arrays.py
```diff
@@ -656,12 +670,12 @@ def validate_and_fill_nuclei(
...
-        'elem': np.array(E),
-        'mass': np.array(mass, dtype=np.float),
-        'real': np.array(real, dtype=np.bool),
-        'elbl': np.array(label)
+        'elea': np.array(A, dtype=np.int, copy=copy),
+        'elez': np.array(Z, dtype=np.int, copy=copy),
+        'elem': np.array(E, copy=copy),
+        'mass': np.array(mass, dtype=np.float, copy=copy),
+        'real': np.array(real, dtype=np.bool, copy=copy),
+        'elbl': np.array(label, copy=copy)
```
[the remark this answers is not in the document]
dgasmith: Yes, but they are lists after the zip so the array will copy the data ;/. No need for copy in this case.

qcelemental/molparse/to_schema.py
```diff
@@ -38,9 +38,17 @@ def to_schema(molrec: Dict, dtype: str, units: str='Bohr', *, np_out: int=False)
...
     else:
         factor = constants.conversion_factor(molrec['units'], units)
-    geom = np.array(molrec['geom']) * factor
+
+
+    geom = np.array(molrec["geom"], copy=copy)
+
+    if factor:
+        geom = geom * factor
+
```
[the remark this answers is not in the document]
dgasmith: Sure, works for me.

qcelemental/molparse/from_schema.py
loriab: My last reservation is that as-is `from_schema` returns mem-independent `geom` if `nfr>1` and mem-dep `geom` if `nfr==1`. If you're ok with that, I'm good with the PR.

dgasmith: We have:
```
    if geom is not None:
        ncgeom = np.array(geom).reshape(-1, 3)
```

I don't see the pathway? Happy to fix.

loriab: two issues:

* nfr=1 returns the raw argument https://github.com/MolSSI/QCElemental/pull/86/files#diff-2d5d131c2b409ebc837aa226ab68e39fR138
whereas https://github.com/MolSSI/QCElemental/pull/86/files#diff-2d5d131c2b409ebc837aa226ab68e39fL149 is surely copying. unless i'm reading it wrong ...

* I think the new block https://github.com/MolSSI/QCElemental/pull/86/files#diff-2d5d131c2b409ebc837aa226ab68e39fR134-R143 needs to move down below the three checks https://github.com/MolSSI/QCElemental/pull/86/files#diff-2d5d131c2b409ebc837aa226ab68e39fR155 because that middle one (despite the frag-focussed message) also catches single-frag [1, 3, 2] cases.

dgasmith: Rewritten slightly to make the first path a shortcut and I believe properly checks for it.

dgasmith:
This is very deep, much better to be safe here.

PR #88, opened 2019-08-09
Port Fractal Autodoc to Elemental, Autodoc Molecule Object

Lnaden:
As it says on the tin.

Probably not ready to merge yet as we may want to work on the Protomodel Metaclass to make this __doc__ tech work automatically rather than as a wrapper function which has to be called on every object.

There definitely should be a review of the descriptions I have affixed to each of the `Molecule` attributes for accuracy. The one which I am the least sure about is the `fix_symmetry` field.

### DIFF
--- a/qcelemental/models/molecule.py
+++ b/qcelemental/models/molecule.py
@@ -9,7 +9,7 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 import numpy as np
-from pydantic import constr, validator
+from pydantic import constr, validator, Schema
 
 from .types import Array
 from .basemodels import ProtoModel
@@ -18,7 +18,7 @@ from ..molparse import from_arrays, from_schema, from_string, to_schema, to_stri
 from ..periodic_table import periodictable
 from ..physical_constants import constants
 from ..testing import compare, compare_values
-from ..util import measure_coordinates, provenance_stamp, which_import
+from ..util import measure_coordinates, provenance_stamp, which_import, auto_gen_docs_on_demand
 
 # Rounding quantities for hashing
 GEOMETRY_NOISE = 8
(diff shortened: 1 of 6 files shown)


qcelemental/models/molecule.py
```diff
@@ -74,44 +74,172 @@ class Config(ProtoModel.Config):
...
+    )
 
     # Molecule data
-    name: str = ""
-    identifiers: Optional[Identifiers] = None
-    comment: Optional[str] = None
-    molecular_charge: float = 0.0
-    molecular_multiplicity: int = 1
+    name: str = Schema(
+        "",
```
loriab: optional, not "", I think

qcelemental/models/molecule.py
```diff
@@ -74,44 +74,172 @@ class Config(ProtoModel.Config):
...
+        description="A common or human-readable name to assign to this molecule. Can be arbitrary."
+    )
+    identifiers: Optional[Identifiers] = Schema(
+        None,
+        description="An optional dictionary of additional identifiers which this Molecule can be referenced by "
```
loriab: "by which this"

qcelemental/models/molecule.py
```diff
@@ -74,44 +74,172 @@ class Config(ProtoModel.Config):
...
-    real: Optional[Array[bool]] = None
-    atom_labels: Optional[Array[str]] = None
-    atomic_numbers: Optional[Array[np.int16]] = None
-    mass_numbers: Optional[Array[np.int16]] = None
+    masses: Optional[Array[float]] = Schema(
+        None,
+        description="An ordered 1-D array-like object of atomic masses of shape (N). Index "
+                    "matches the index of all other per-atom settings like ``symbols`` and ``real``. If this is not "
+                    "provided, the mass of each atom is inferred from their most common isotope. If this is provided,"
+                    "every atom must have its mass specified, even if they are the common isotopes."
```
loriab: * missing `[u]`
* "not provided, the mass of the most abundant isotope is used."
* the validator can handle `[None, 2.014, None]` afaik
* "A 1-D array-like object of atomic masses [u] of shape (nat). Atom order in array-like matches other per-atom fields like `symbols` and `geometry`."

qcelemental/models/molecule.py
dgasmith: Yup.

qcelemental/models/molecule.py
Lnaden: > the validator can handle [None, 2.014, None] afaik

I don't see anywhere that it would not accept that, but I also don't see anywhere which would convert the `None` inside the list to the correct mass. The only auto-mass assignment I can see are these two lines:

```python
        if values["masses"] is None:  # Setup masses before fixing the orientation
            values["masses"] = np.array([periodictable.to_mass(x) for x in values["symbols"]])
```

which would only trigger if the `masses` field is None as a whole, not a list which has None entries.

qcelemental/models/molecule.py
Lnaden: This has been a non-optional string before now which just defaulted to the empty string.. I made it optional, we'll see if it breaks anything

qcelemental/models/molecule.py
```diff
@@ -74,44 +74,172 @@ class Config(ProtoModel.Config):
 
 
 class Molecule(ProtoModel):
+    """
+    The fundamental representation of a Molecule inside the QCArchive Ecosystem. This model contains
+    data for atomic elements, coordinates, connectivity, charges, fragmentation, etc. The model also
+    supports orientation and measurement utilities.
+
+    All Molecule objects are oriented via inertia tensor with the center-of-mass at (0,0,0) unless
```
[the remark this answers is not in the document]
Lnaden: There is no good way to f-string the docstring, so I hard wrote the values (as they stand) and gave pointer to the variables if people wanted to look at them. Let me know what you think of the wording.

qcelemental/models/molecule.py
dgasmith:
```
>>> import numpy as np
>>> np.array([4, 3]) is None
False
```

Lnaden: Yes, but I think the point was that if you pass in something like 

```
symbols=["C", "C", "C"]
masses=[None, 12.011, None]
```

at which point are the `None` in the List converted to the correct mass based on symbol? The only checks I can see are:

* Is the masses object None?
* Does the masses List shape equal that of the symbols list?

qcelemental/models/molecule.py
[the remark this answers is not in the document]
loriab: Ah, ok, thanks for the extra sentence. variable formatting for GEOMETRY NOISE, etc. probably sufficient w/o extra "Truncation ... value." sentence.

qcelemental/models/molecule.py
loriab: the `to_schema(from_schema())` should pass most everything that doesn't have validation turned off through `molparse.from_arrays()` which can handle the `[None, 12.011, None]` case. haven't explicitly tried it, though.

dgasmith:
```
>>> [None, 12.011, None] is None
False
```

None is not checking the elements, it only checking the *object*.

Lnaden: > None is not checking the elements, it only checking the object.

Thats not what I have concerns about, in the original point Lori made

> the validator can handle [None, 2.014, None] afaik

Which I agree with, the validator *can* handle that and will accept that as a valid input, so long as the length of masses matches the length of symbols. My concern though however is that those `None` entries in the list *are accepted/not converted in the first place*, resulting in the masses having undefined/None entries on some of the atoms.

I think there will be a problem, for instance, if we are orienting the molecule and there are `None` in the masses list, when we get to the line `new_geometry -= np.average(new_geometry, axis=0, weights=np_mass)`, then it will throw a type error.

Lnaden: Thinking about it some more, the type coercion on `Optional[Array[float]]` won't allow things like `[None 1.2345, None]`, so this is likely a moot point and the validator will reject such energies.

qcelemental/models/molecule.py
```diff
@@ -186,25 +194,27 @@ class Molecule(ProtoModel):
         )
     fragments: Optional[List[Array[np.int32]]] = Schema(
```
loriab: I thought @dgasmith ended up with fragment* fields remaining lists, not converting to ndarray?

qcelemental/models/molecule.py
```diff
@@ -186,25 +194,27 @@ class Molecule(ProtoModel):
...
         None,
-        description="An indication of which sets of atoms are fragments within the Molecule. This is a list "
-                    "of 1-D array-like objects. Each entry in the list indicates a new fragment. The index "
-                    "of this list matches the index of ``fragment_charges`` and ``fragment_multiplicities``. "
+        description="An indication of which sets of atoms are fragments within the Molecule. This is a list of shape "
+                    "(NFrag) of 1-D array-like objects of arbitrary shape. Each entry in the list indicates a new "
```
loriab: * arbitrary length, not shape
* better explicitly say "0-indexed" atoms whenever either could be true.

qcelemental/models/molecule.py
```diff
@@ -105,26 +110,28 @@ class Molecule(ProtoModel):
         ...,
         # LNN Note: "Ordered" might not be the least ambiguous term
         # (e.g. "ordered" as in linked ordering or "ordered" as in alphabetical?)
-        description="An ordered 1-D array-like object of atomic elemental symbols of shape (N). Index of this list "
+        description="An ordered 1-D array-like object of atomic elemental symbols of shape (NAtom). Index of this list "
```
loriab: "An ordered `(nat,)` array-like" says the same thing more compactly, if you care for it.

qcelemental/models/molecule.py
```diff
@@ -186,25 +194,27 @@ class Molecule(ProtoModel):
...
                     "the fragment. The atom indices match the index of all other per-atom settings like ``symbols`` "
                     "and ``real``."
     )
     fragment_charges: Optional[List[float]] = Schema(
         None,
-        description="The total charge of each fragment in the ``fragments`` list. The index of this list matches the "
-                    "index of ``fragment`` list. This must be provided if the ``molecular_charge`` is not 0, otherwise "
-                    "it is assumed fragments are all 0 charge as well (only a condition if there are ``fragments``)."
+        description="The total charge of each fragment in the ``fragments`` list of shape (NFrag). The index of this "
+                    "list matches the index of ``fragment`` list. This must be provided if the ``molecular_charge`` is "
```
loriab: no on the "must be provided" here and below (and maybe on molecular_chg/mult?). Validation will fill in frag chgmult and switch mult to open shell if the system demands it according to [rules](https://github.com/MolSSI/QCElemental/blob/master/qcelemental/molparse/chgmult.py#L103-L134)

qcelemental/models/molecule.py
```diff
@@ -105,26 +110,28 @@ class Molecule(ProtoModel):
...
     )
     geometry: Array[float] = Schema(
         ...,
-        description="An ordered 2-D array-like object of shape (N,3) for atomic coordinates. Index of the "
-                    "first dimension matches the index of all other per-atom settings like ``symbols`` and ``real``."
+        description="An ordered 2-D array-like object of shape (NAtom,3) for XYZ atomic coordinates [a0]. "
+                    "Atom ordering is fixed; that is, a consumer who shuffles atoms must not reattach the input "
+                    "(pre-shuffling) molecule schema instance to any output (post-shuffling) per-atom results "
+                    "(e.g., gradient). Index of the first dimension matches the index of all other per-atom settings "
+                    "like ``symbols`` and ``real``."
```
loriab: I think we may be conflating documentation targets. There's internal storage (ndarray `(nat, 3)`). There's what can be input to the object constructor (array-like that somehow total 3*nat). There's serialization format (`(3*nat, )` if msgpack unavailable, if I understand the new stuff correctly). Should we be clear about target audience context?

qcelemental/models/molecule.py
Lnaden: I'll take compact, do you have strong opinion on `NAtom` vs `Nat` or `nat` for the shape names? I'm of the opinion that the doc string (and in turn the docs on the page) should be more human-consumable and verbose instead of `nat`, `nfr`, `nvib`, `ndof` which takes some level of contextual or experience to remember what they all mean.

qcelemental/models/molecule.py
Lnaden: Good question. I think the main user who will interact with these docs are those which will actually either be constructing these objects themselves (`Molecule(geometries=...)`) or those who are working with the constructed objects (`mol.geometries[1,:]`) , so I think targeting those users are the ones we should worry about rather than the sterilization crowd. 

I do agree there is a bit ambiguity because someone can feed in a 1-D array-like of length `3*NAtom` and it will be converted to the `(NAtom, 3)` shape. I can clean that part up if you think that will help

qcelemental/models/molecule.py
loriab: I do have strong feelings and hate mixed case, but I won't impose them on docstrings, only the code. :-)

qcelemental/models/molecule.py
dgasmith: This signature is correct, I gave up and went with it.

qcelemental/models/molecule.py
Lnaden: I'd rather it be consistent everywhere. I'll go through and lcase them all to match the variables in the code. Hopefully this will be help both end-users and devs.

qcelemental/models/molecule.py
loriab: Maybe `geometry` only gets a shout-out to the serialization crowd b/c that was (strangely) a main bullet result of the MolSSI schema workshop.

But, yes, I think outright stating the internal storage in first sentence/paragraph, then in second paragraph expanding on allowed init values is a good separation. It'll be especially handy for chgmult, where internal always has a value, while upon init you can give all, nothing, or in between.

PR #116, opened 2019-08-19
Update master docs for 0.6.1

Lnaden:
This keeps the change log in sync on 1 location. I can't just make a
PR from the cherry-pick maintenance branch since it tries to make the
cherry-pick change on top of it.

Still working out order-of-operations. I think it may be good to make the doc change in `master`, and cherry-pick the docs as well as the code.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -13,6 +13,15 @@ Changelog
 .. Bug Fixes
 .. +++++++++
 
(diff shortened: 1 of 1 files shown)


dgasmith:
Hard to say, I wonder if docs only go into master?

Lnaden:
Thats probably the simplest going forward. I realized I had made the docs in the 0.6.x branch and that master was missing them here.

PR #118, opened 2019-08-20
Models Repr

dgasmith:
This PR attempts to give reasonable string representation for all models and ensure that skip_defaults are called on certain models to prevent the alteration of defaulted fields in downstream Pydantic uses.

### DIFF
--- a/qcelemental/models/procedures.py
+++ b/qcelemental/models/procedures.py
@@ -36,6 +36,9 @@ class OptimizationInput(ProtoModel):
 
     provenance: Provenance = provenance_stamp(__name__)
 
+    def __str__(self):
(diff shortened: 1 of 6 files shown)


dgasmith:
Rebased to try to trigger Codecov again.

qcelemental/models/procedures.py
```diff
@@ -36,6 +36,9 @@ class OptimizationInput(ProtoModel):
 
     provenance: Provenance = provenance_stamp(__name__)
 
+    def __str__(self) -> str:
+        return f"{self.__class__.__name__}(model='{self.input_specification.model.dict()}' molecule_hash='{self.initial_molecule.get_hash()[:7]}')"
```
Lnaden: How long will this one get to be?

dgasmith: Usually sub 60 characters?

PR #122, opened 2019-08-23
LGTM Cleanup Pass

Lnaden:
Most of these are test related LGTM cleanups. There are 2 however I want to call attention to (and will do so in a review)

### DIFF
--- a/qcelemental/testing.py
+++ b/qcelemental/testing.py
@@ -111,7 +111,7 @@ def compare_values(expected,
     if xptd.shape != cptd.shape:
         return return_handler(False, label,
                               f"""\t{label}: computed shape ({cptd.shape}) does not match ({xptd.shape}).""",
-                              return_message, quiet)
+                              return_message, quiet)  # lgtm: [py/syntax-error]
 
     digits1 = abs(int(np.log10(atol))) + 2
     digits_str = f'to atol={atol}'
--- a/qcelemental/tests/test_importing.py
+++ b/qcelemental/tests/test_importing.py
@@ -73,20 +73,6 @@ def test_which_f_raisemsg():
     assert str(e.value).endswith("Command 'evills' not found in envvar PATH. Install `evills`.")
 
 
-def test_which_f_raise():
-    with pytest.raises(ModuleNotFoundError) as e:
-        qcel.util.which('evills', raise_error=True)
-
-    assert str(e.value).endswith("Command 'evills' not found in envvar PATH.")
-
-
-def test_which_f_raisemsg():
-    with pytest.raises(ModuleNotFoundError) as e:
-        qcel.util.which('evills', raise_error=True, raise_msg='Install `evills`.')
-
-    assert str(e.value).endswith("Command 'evills' not found in envvar PATH. Install `evills`.")
-
-
 def test_safe_version():
     assert 'v' + qcel.util.safe_version(qcel.__version__) == qcel.__version__
 
--- a/qcelemental/tests/test_molecule.py
+++ b/qcelemental/tests/test_molecule.py
@@ -364,6 +364,8 @@ def test_get_fragment(group_fragments, orient):
         assert dimers[0].get_hash() == dimers[3].get_hash()
         assert dimers[1].get_hash() == dimers[4].get_hash()
         assert dimers[2].get_hash() == dimers[5].get_hash()
+    else:
+        assert 0  # lgtm: [py/unreachable-statement]
 
     ghdimers_nelectrons = [2, 2, 10, 10, 10, 10]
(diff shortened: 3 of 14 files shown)


qcelemental/testing.py
```diff
@@ -111,7 +111,7 @@ def compare_values(expected,
     if xptd.shape != cptd.shape:
         return return_handler(False, label,
                               f"""\t{label}: computed shape ({cptd.shape}) does not match ({xptd.shape}).""",
-                              return_message, quiet)
+                              return_message, quiet)  # lgtm: [py/syntax-error]
```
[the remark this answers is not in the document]
loriab: I figured b/c f-string

qcelemental/tests/test_importing.py
```diff
@@ -73,20 +73,6 @@ def test_which_f_raisemsg():
...
-        qcel.util.which('evills', raise_error=True)
-
-    assert str(e.value).endswith("Command 'evills' not found in envvar PATH.")
-
-
-def test_which_f_raisemsg():
-    with pytest.raises(ModuleNotFoundError) as e:
-        qcel.util.which('evills', raise_error=True, raise_msg='Install `evills`.')
-
-    assert str(e.value).endswith("Command 'evills' not found in envvar PATH. Install `evills`.")
```
loriab: what's wrong with these tests?

qcelemental/tests/test_molparse_pubchem.py
```diff
@@ -38,12 +38,12 @@
         0.000000, -2.139300, 1.256400, 0.000100, -0.018400, -2.480900, -0.000100, 0.018400, 2.480800, 0.000000,
         2.139400, -1.256300, 0.000100, 2.157700, 1.224500, 0.000000
     ]),
-    'elbl': np.array(['C', 'C', 'C', 'C', 'C', 'C', 'H', 'H', 'H', 'H', 'H', 'H']),
     'elea': np.array([12, 12, 12, 12, 12, 12, 1, 1, 1, 1, 1, 1]),
     'elez': np.array([6, 6, 6, 6, 6, 6, 1, 1, 1, 1, 1, 1]),
     'elem': np.array(['C', 'C', 'C', 'C', 'C', 'C', 'H', 'H', 'H', 'H', 'H', 'H']),
     'mass': np.array([12., 12., 12., 12., 12., 12., 1.00782503, 1.00782503, 1.00782503, 1.00782503, 1.00782503, 1.00782503]),
     'real': np.array([True, True, True, True, True, True, True, True, True, True, True, True]),
+    # 'elbl': np.array(['C', 'C', 'C', 'C', 'C', 'C', 'H', 'H', 'H', 'H', 'H', 'H']),
```
loriab: there's two definitions of `elbl` -- sort of a pre-parse and post-parse/storage. this is probably a residual from what I needed for testing earlier. can remove the "filled" line.

qcelemental/tests/test_molecule.py
```diff
@@ -364,8 +364,6 @@ def test_get_fragment(group_fragments, orient):
         assert dimers[0].get_hash() == dimers[3].get_hash()
         assert dimers[1].get_hash() == dimers[4].get_hash()
         assert dimers[2].get_hash() == dimers[5].get_hash()
-    else:
-        assert 0
```
loriab: these are a check that if you don't do one of the known/coded combos, everything explodes. not strictly necessary, but I don't think they hurt.

qcelemental/testing.py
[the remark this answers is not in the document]
Lnaden: But f-strings work fine in 3.6?

qcelemental/tests/test_importing.py
Lnaden: Duplicates of the ones that were above them.

https://github.com/MolSSI/QCElemental/blob/9f3bbf0f4614e1b66ddd7e8df74ccb7be5a72ac3/qcelemental/tests/test_importing.py#L62-L71

qcelemental/testing.py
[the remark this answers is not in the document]
loriab: yes, but all lgtm knows is that your code is py3 compatible, not py36. agreed that it shouldn't be triggering a lot more if my theory holds, though.

qcelemental/tests/test_molecule.py
Lnaden: True, but in our current testing was unreachable because the Pytest parameterization was either `True, True` or `False, False`. So either LGTM caught that or..... it was a fluke. I can add it back in though for safety and we just say "meh" to the LGTM warning here.

qcelemental/testing.py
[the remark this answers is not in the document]
Lnaden: > be triggering a lot more if my theory holds, though

That would be my guess, but it explicitly stated it was a 3.6 compatibility issue. I'm guessing false positive and I silenced it.

PR #123, opened 2019-08-29
MyPy, lots of MyPy

Lnaden:
This PR makes Elemental pass MyPy. I've done what I can to make not
use the `type: ignore` comment best I can, but Pydantic and MyPy don't
seem to play nice in many cases.

Outstanding issues I have no good solution to:
* Typed Dict not available until 3.8
* `lambda` functions with kwargs that have defaults can't be inffered and Callable does not work.
* MyPy does not like the `constr` from Pydantic
* Not sure how to fix the Array type acceptance, if I try to make it have a `Generic[T]` type, it complains about metaclass types
* Not sure if the `Array` Generic type is really a good idea?
* `compare_X` functions cannot import ProtoModel since it creates a recursive import as other things import from testing.

Next major step would be to include the `.pyi` stub files so things like Engine can recognize the typed functions


Lnaden:
One thing I found was that the `testing` module is actually imported by other places in the code to gain access to the `compare_X` routines. What are the thoughts on porting those to the `util` module so I can try to do proper typing without a recursive import? The issue with that is the `testing` module then becomes depleted of most its functions.

I could also just leave it and rely on forward ref type hints for free.

loriab:
I'm not keen on emigration from `testing` b/c I've got several other projects importing `qcel.testing` for its goodies. If making `testing` into a directory and separating pieces that way helped typing, I'm fine with that.

Lnaden:
>  If making testing into a directory and separating pieces that way helped typing, I'm fine with that.

Eh... maybe? It would take a bit of engineering. I'm fine leaving it as is and suspected the `testing` module was being used elsewhere due to its handy nature, hence why I asked. I'll just leave it as a forward ref.

PR #129, opened 2019-09-09
Molpro: Writing the multiplicity

sjrl:
- Changed how multiplicity is set in the Molpro input file so it is actually used by the program. 
- The previous way of setting was never used by the program (a default guess was used instead)


dgasmith:
Will this break QCEngine where we compare exact strings text for the input formatters?

sjrl:
Yes this will. I'll update the files in QCEngineRecords to reflect these changes.

PR #133, opened 2019-09-19
Multipoles

loriab:


qcelemental/models/multipole.py
```diff
@@ -0,0 +1,50 @@
...
+from .types import Array
+
+
+class Singlepole(ProtoModel):
+    """Model for charges, dipoles, or quadrupoles, etc."""
+
+    angular_momentum: int = Schema(..., description="Angular momentum for this singlepole.")
+    exponents: List[float] = Schema(..., description="Exponents for this singlepole.")
+    geometry: Array[float] = Schema(..., description="Location of the singlepoles.")
+    coefficients: List[List[float]] = Schema(..., description="Coefficients for each AM component. Psi4 AM ordering convention: https://github.com/evaleev/libint/blob/5458ab2fb2fd51dcd19f1e8122a451f2a0808074/doc/progman/progman.tex#L1006-L1008")
```
dgasmith: This is the CCA standard, we should probably avoid Psi4 in schema objects.

Is there such a thing as spherical vs cartesian multipoles?

qcelemental/models/multipole.py
```diff
@@ -0,0 +1,50 @@
...
+        for ipts in v:
+            if len(ipts) != ncoeff:
+                raise ValueError(f"The length of coefficients AM does not match the AM. {len(ipts)} != {coeff}")
+
+        return v
+
+
+class Multipoles(ProtoModel):
+
+    poles: List[Singlepole] = []
```
dgasmith: The idea being here that we would have a point charge `Singlepole` that would contain a list of centers and then another `dipole` with another list of centers?

@andysim do you anticipate the centers to overlap between charges/dipoles/etc?

In general, what order of magnitude range do we expect in the number of these elements?

loriab: Yes on your first question. Probably Multipoles gets the fancy `__init__` that can broadcast exponents up to a fixed AM, etc.

PR #137, opened 2019-09-24
Turbomole

eljost:
This PR adds the ability to generate coordinate strings in the Turbomole format. I also added a corresponding test in test_molparse_to_string.py

### DIFF
--- a/qcelemental/molparse/to_string.py
+++ b/qcelemental/molparse/to_string.py
@@ -74,7 +74,8 @@ def to_string(molrec: Dict,
         "molpro": "Bohr",
         "nwchem": "Bohr",
         "psi4": "Bohr",
-        "terachem": "Bohr"
+        "terachem": "Bohr",
+        "turbomole": "Bohr",
     }
     if dtype not in default_units:
         raise KeyError(f"dtype '{dtype}' not understood.")
@@ -289,6 +290,13 @@ def to_string(molrec: Dict,
         ])
         data.keywords = {}
 
+    elif dtype == 'turbomole':
+        # In Turbomole coord files the coordinates come first, and the atomic
+        # symbol comes afterwards.
+        coords3d = molrec["geom"].reshape(-1, 3)
+        smol = [f"{x: .8f} {y: .8f} {z: .8f} {atom.lower()}"
+                for (x, y, z), atom in zip(coords3d, molrec["elem"])]
+        smol = ["$coord"] + smol + ["$end"]
     else:
         raise KeyError(f"dtype '{dtype}' not understood.")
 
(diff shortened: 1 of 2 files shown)


qcelemental/molparse/to_string.py
```diff
@@ -289,6 +290,14 @@ def to_dict(self) -> Dict:
...
         data.keywords = {}
 
+    elif dtype == 'turbomole':
+        # In Turbomole coord files the coordinates come first, and the atomic
+        # symbol comes afterwards.
+        # Handling of ghost atoms is done in the basis section of the control
+        # file by setting the nuclear charge of certain atoms to zero.
+        smol = [f"{x: >20.14f}{y: >20.14f}{z: >20.14f}{atom.lower():>8s}"
+                for (x, y, z), atom in zip(geom, molrec["elem"])]
+        smol = ["$coord"] + smol + ["$end"]
```
loriab: Does turbomole require the 20.14f and 8s formats? And does it permit angstrom input?

eljost: I'm using a similar formatting in my code (https://github.com/eljost/pysisyphus/blob/dev/pysisyphus/calculators/Calculator.py#L217) and it works ans closely resembles the files the Turbomole utility x2t produces.

Regarding the angstrom input the manual says: "Molecules and their structures are specified by coordinates of its atoms, within the program invariably by Cartesian coordinates in atomic units (Ångstrøm would also do)" but I don't know how to tell TM to use Angstrom, and the manual doesn't tell either. So I guess TM requires coordinates in atomic units.

loriab: your analysis of the `_atoms_formatter` is spot on, but I can adjust it for turbomole. how sensible of it to accept only Bohr (in practice).

Maybe this is the angstrom route? I'm happy to continue ignoring it.
```
a <file>         : ADD ATOMIC COORDINATES FROM FILE <file>
aa <file>        : ADD ATOMIC COORDINATES IN ANGSTROEM UNITS FROM FILE <file
```

eljost: Ok, this seems like it ... do you think Angstrom should be supported somehow in the TurbomoleHarness? In the end it shouldn't matter. In to_string the coordinates get converted to the right unit (to Bohr), then I would hardcode the "a" route in the TurbomoleHarness.

loriab: No need, imo. Bohr is the way we want it conveyed for qcng. For someone using `to_string` independent of qcng, if the bohr/ang could have been conveyed in the string, it would have been worth adding. But since it's external with a/aa, best to leave it out.

PR #138, opened 2019-09-24
handle turbomole case where element at end of line

loriab:
minor modifications of #137 so that the format commands work for turbomole, too


dgasmith:
@eljost please review.

PR #139, opened 2019-09-26
Wavefunction Model

dgasmith:
Adds a validated wavefunction property attribute to the Result Model. This matched the MolSSI QCSchema.

### DIFF
--- a/qcelemental/models/basis.py
+++ b/qcelemental/models/basis.py
@@ -37,6 +37,21 @@ class ElectronShell(ProtoModel):
 
         return v
 
+    def nfunctions(self) -> int:
+        """
+        Computes the number of basis functions on this shell.
+
+        Returns
+        -------
+        int
+            The number of basis functions on this shell.
+        """
+
+        if self.harmonic_type == 'spherical':
+            return sum((2 * L + 1) for L in self.angular_momentum)
+        else:
+            return sum(((L + 1) * (L + 2) // 2) for L in self.angular_momentum)
+
 
 class ECPType(str, Enum):
     """
@@ -110,3 +125,23 @@ class BasisSet(ProtoModel):
             raise ValueError(f"'atom_map' contains unknown keys to 'center_data': {missing}.")
 
         return v
+
+    def nbf(self) -> int:
+        """
+        Number of basis functions in the basis set.
+
+        Returns
+        -------
+        int
+            The number of basis functions.
+        """
+
+        center_count = {}
+        for k, center in self.center_data.items():
+            center_count[k] = sum(x.nfunctions() for x in center.electron_shells)
+
+        ret = 0
+        for center in self.atom_map:
+            ret += center_count[center]
+
+        return ret
--- a/qcelemental/models/basis.py
+++ b/qcelemental/models/basis.py
@@ -99,14 +99,14 @@ class BasisSet(ProtoModel):
     description: Optional[str] = Schema(None, description="A brief description of the basis set.")
     center_data: Dict[str, BasisCenter] = Schema(..., description="A mapping of all types of centers available.")
     atom_map: List[str] = Schema(
-        ..., description="Mapping of all centers in the parent molecule to centers in `basis_data`.")
+        ..., description="Mapping of all centers in the parent molecule to centers in `center_data`.")
 
-    @validator('basis_atom_map', whole=True)
+    @validator('atom_map', whole=True)
     def _check_atom_map(cls, v, values):
         sv = set(v)
-        missing = sv - values["basis_data"].keys()
+        missing = sv - values["center_data"].keys()
 
         if missing:
-            raise ValueError(f"'basis_atom_map' contains unknown keys to 'basis_data': {missing}.")
+            raise ValueError(f"'atom_map' contains unknown keys to 'center_data': {missing}.")
 
         return v
--- a/qcelemental/models/basis.py
+++ b/qcelemental/models/basis.py
@@ -116,6 +116,8 @@ class BasisSet(ProtoModel):
     atom_map: List[str] = Schema(
         ..., description="Mapping of all centers in the parent molecule to centers in `center_data`.")
 
+    nbf: Optional[int] = Schema(None, description="The number of basis functions.")
+
     @validator('atom_map', whole=True)
     def _check_atom_map(cls, v, values):
         sv = set(v)
@@ -126,7 +128,25 @@ class BasisSet(ProtoModel):
 
         return v
 
-    def nbf(self) -> int:
+    @validator('nbf', always=True)
+    def _check_nbf(cls, v, values):
+
+        # Bad construction, pass on errors
+        try:
+            nbf = cls._calculate_nbf(values["atom_map"], values["center_data"])
+        except KeyError:
+            return v
+
+        if v is None:
+            v = nbf
+        else:
+            if v != nbf:
+                raise ValidationError("Calculated nbf does not match supplied nbf.")
+
+        return v
+
+    @classmethod
+    def _calculate_nbf(self, atom_map, center_data) -> int:
         """
         Number of basis functions in the basis set.
 
@@ -137,11 +157,11 @@ class BasisSet(ProtoModel):
         """
 
         center_count = {}
-        for k, center in self.center_data.items():
+        for k, center in center_data.items():
             center_count[k] = sum(x.nfunctions() for x in center.electron_shells)
 
         ret = 0
-        for center in self.atom_map:
+        for center in atom_map:
             ret += center_count[center]
 
         return ret
--- a/qcelemental/models/basis.py
+++ b/qcelemental/models/basis.py
@@ -95,10 +95,10 @@ class BasisSet(ProtoModel):
     schema_name: constr(strip_whitespace=True, regex="qcschema_basis") = "qcschema_basis"
     schema_version: int = 1
 
-    basis_name: str = Schema(..., description="A standard basis name if available (e.g., 'cc-pVDZ'.")
+    name: str = Schema(..., description="A standard basis name if available (e.g., 'cc-pVDZ'.")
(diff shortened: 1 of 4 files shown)


qcelemental/models/results.py
```diff
@@ -97,6 +98,107 @@ def __str__(self):
...
+    scf_orbitals_b: Optional[Array[float]] = Schema(None, description="SCF beta-spin orbitals.")
+    scf_density_a: Optional[Array[float]] = Schema(None, description="SCF alpha-spin density matrix.")
+    scf_density_b: Optional[Array[float]] = Schema(None, description="SCF beta-spin density matrix.")
+    scf_fock_a: Optional[Array[float]] = Schema(None, description="SCF alpha-spin Fock matrix.")
+    scf_fock_b: Optional[Array[float]] = Schema(None, description="SCF beta-spin Fock matrix.")
+    scf_eigenvalues_a: Optional[Array[float]] = Schema(None, description="SCF alpha-spin eigenvalues.")
+    scf_eigenvalues_b: Optional[Array[float]] = Schema(None, description="SCF beta-spin eigenvalues.")
+    scf_occupations_a: Optional[Array[float]] = Schema(None, description="SCF alpha-spin occupations.")
+    scf_occupations_b: Optional[Array[float]] = Schema(None, description="SCF beta-spin occupations.")
+
```
mattwelborn: Add localized orbital results.

dgasmith: I would like to skip this on the first pass. I am not quite sure I understand the exact requirements there.

qcelemental/models/results.py
```diff
@@ -97,6 +98,107 @@ def __str__(self):
...
+    def _assert1d(cls, v, values):
+        bas = values.get("basis", None)
+
+        # Do not raise multiple errors
+        if bas is None:
+            return v
+
+        try:
+            v = v.reshape(bas.nbf)
+        except (ValueError, AttributeError):
```
mattwelborn: Two things: 

1. ValueError message should be "shape (nbf)'. 
2. is `nbf` guaranteed to be `nmo` for SCF, or are there edge cases?

qcelemental/models/results.py
```diff
@@ -97,6 +98,107 @@ def __str__(self):
...
+        bas = values.get("basis", None)
+
+        # Do not raise multiple errors
+        if bas is None:
+            return v
+
+        try:
+            v = v.reshape(bas.nbf, bas.nbf)
+        except (ValueError, AttributeError):
+            raise ValueError("Matrix must be castable to shape (nbf, nbf)!")
```
mattwelborn: Same comment as above, `(nbf, nbf) ?= (nbf, nmo)` for SCF quantities.

qcelemental/models/results.py
dgasmith: Could point, I made a new validator for `(nbf, -1)` for now

qcelemental/models/results.py
```diff
@@ -97,6 +98,107 @@ def __str__(self):
...
+        return v
+
+    @validator('orbitals_a',
+               'orbitals_a',
+               'density_a',
+               'density_b',
+               'fock_a',
+               'fock_b',
+               'eigenvalues_a',
+               'eigenvalues_b',
```
mattwelborn: add occupations

qcelemental/tests/test_models_results.py
```diff
@@ -86,22 +91,84 @@ def test_basis_electron_center_raises():
...
-                          basis_atom_map=["something_odd"])
+        assert basis.BasisSet(name="custom_basis", center_data=center_data, atom_map=["something_odd"])
+
+
+@pytest.fixture(scope="function")
+def wavefunction_data_fixture():
+    bas = basis.BasisSet(name="custom_basis",
+                         center_data=center_data,
+                         atom_map=["bs_sto3g_o", "bs_sto3g_h", "bs_sto3g_h"])
+    c_matrix = np.random.rand(bas.nbf, bas.nbf)
```
mattwelborn: This might be a good place to test `nao != nmo`

qcelemental/models/results.py
dgasmith: This gets tricky as then `nbf` may change *again* after SCF, after MP2, and even after CCSD...

qcelemental/tests/test_models_results.py
```diff
@@ -86,22 +91,84 @@ def test_basis_electron_center_raises():
...
+        "molecule": mol,
+        "driver": "energy",
+        "model": {
+            "method": "UFF"
+        },
+        "return_result": 5,
+        "wavefunction": {
+            "basis": bas,
+            "scf_orbitals_a": c_matrix,
+            "orbitals_a": "scf_orbitals_a"
```
mattwelborn: for restricted closed-shell results, should it be convention for beta components to always be reported in addition to alpha ones, especially for return results? This would match the way we handle `calcinfo_nalpha` and `calcinfo_nbeta`.

dgasmith: I think we will have to leave it sparse, a factor of two in common cases is hard to give up for these large quantities.

qcelemental/models/results.py
mattwelborn: I understand how nmo would change, but why would nbf change?

qcelemental/tests/test_models_results.py
mattwelborn: I think I'm okay with this, but we should document it somewhere (maybe in the protocols).

qcelemental/models/results.py
dgasmith: Sorry, replace above `nbf` with `nmo`.

dgasmith:
Ok, changes up and patched. Would prefer to merge this now unless there are blockers and open a new protocols PR which will need further discussion.

qcelemental/models/basis.py
```diff
@@ -37,6 +37,21 @@ def _check_coefficient_length(cls, v, values):
 
         return v
 
+    def nfunctions(self) -> int:
```
loriab: I know you're against abbreviations, but `nbf` seems more descriptive

qcelemental/models/results.py
mattwelborn: I think this okay; we decided that nmo was allowed to be anything and that we wouldn't validate it.

qcelemental/models/basis.py
```diff
@@ -95,18 +110,58 @@ class BasisSet(ProtoModel):
...
+        try:
+            nbf = cls._calculate_nbf(values["atom_map"], values["center_data"])
+        except KeyError:
+            return v
+
+        if v is None:
+            v = nbf
+        else:
+            if v != nbf:
+                raise ValidationError("Calculated nbf does not match supplied nbf.")
```
loriab: should user _ever_ be supplying nbf? I'd think it'd always be on-the-fly. Or is this for a revalidation check dumping in a pre-built object that's been living loose as a dictionary for a while?

qcelemental/models/results.py
```diff
@@ -97,6 +98,119 @@ def __str__(self):
...
+
+        try:
+            v = v.reshape(-1)
+        except (ValueError, AttributeError):
+            raise ValueError("Vector must be castable to shape (-1, )!")
+        return v
+
+    @validator(
+        'scf_orbitals_a',
+        'scf_orbitals_a',
```
loriab: `_b`?

qcelemental/models/results.py
```diff
@@ -97,6 +98,119 @@ def __str__(self):
...
+               'fock_b',
+               'eigenvalues_a',
+               'eigenvalues_b',
+               'occupations_a',
+               'occupations_b',
+               whole=True)
+    def _assert_exists(cls, v, values):
+
+        if values.get(v, None) is None:
+            raise ValueError(f"Return quantity {v} does not exists in the values.")
```
loriab: singular exists, then adjust test case, too.

qcelemental/models/basis.py
```diff
@@ -95,18 +110,58 @@ class BasisSet(ProtoModel):
...
+    name: str = Schema(..., description="A standard basis name if available (e.g., 'cc-pVDZ'.")
     description: Optional[str] = Schema(None, description="A brief description of the basis set.")
-    basis_data: Dict[str, BasisCenter] = Schema(..., description="A mapping of all types of centers available.")
-    basis_atom_map: List[str] = Schema(
-        ..., description="Mapping of all centers in the parent molecule to centers in `basis_data`.")
+    center_data: Dict[str, BasisCenter] = Schema(..., description="A mapping of all types of centers available.")
+    atom_map: List[str] = Schema(
+        ..., description="Mapping of all centers in the parent molecule to centers in `center_data`.")
+
+    nbf: Optional[int] = Schema(None, description="The number of basis functions.")
```
loriab: should BasisSet schema_version bump to 2 with these renamings so that easily differentiate from Ben's records?

qcelemental/models/basis.py
dgasmith: On a single internal shell `nfunctions` seem more appropriate? I could go either way, just a shell isn't a full basis.

qcelemental/models/basis.py
dgasmith: It is hard to not allow a variable in Pydantic, but ensure it is always a value. This is my awkward solution to this problem.

qcelemental/models/basis.py
dgasmith: I don't know if the BasisSet schema was ever fully released. It would agree in principle however.

qcelemental/models/basis.py
loriab: good call. I'm back to `nfunctions` in this context

qcelemental/models/basis.py
loriab: ah, you want `bsinstance.nbf` valid rather than `bsinstance.nbf()` for otf? sounds reasonable

PR #140, opened 2019-09-26
Protocols

dgasmith:
This PR starts protocols for discussion. Items:
1) Should protocols be complex? Right now we just trim data, but could they be used to specify that we require only the final orbitals in an optimization?
2) Should the protocols automatically trim data without warnings?


qcelemental/tests/test_models_results.py
```diff
@@ -65,6 +65,69 @@
...
+    }
+
+
+@pytest.fixture(scope="function")
+def wavefunction_data_fixture(result_data_fixture):
+    bas = basis.BasisSet(name="custom_basis",
+                         center_data=center_data,
+                         atom_map=["bs_sto3g_o", "bs_sto3g_h", "bs_sto3g_h"])
+    c_matrix = np.random.rand(bas.nbf, bas.nbf)
+    result_data_fixture["wavefunction"] = {"basis": bas, "scf_orbitals_a": c_matrix, "orbitals_a": "scf_orbitals_a"}
```
loriab: shouldn't "orbitals_a" fail to validate as an array with a string on rhs?

qcelemental/tests/test_models_results.py
```diff
@@ -163,3 +194,21 @@ def test_wavefunction_return_result_pointer(wavefunction_data_fixture):
...
+])
+def test_optimization_trajectory_protocol(keep, indices, optimization_data_fixture):
+
+    if keep is not None:
+        optimization_data_fixture["protocols"] = {"keep_trajectory": keep}
+    opt = qcel.models.Optimization(**optimization_data_fixture)
+
+    assert len(opt.trajectory) == len(indices)
+    for result, index in zip(opt.trajectory, indices):
+        assert result.return_result == index
```
loriab: isn't lhs an energy and rhs an integer?

qcelemental/models/results.py
```diff
@@ -211,6 +209,24 @@ def _assert_exists(cls, v, values):
         return v
 
 
+class KeepWavefunctionProtocolEnum(str, Enum):
+    """
+    Wavefunction to keep from a Result computation.
+    """
+    all = "all"
```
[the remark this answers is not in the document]
dgasmith: `whole`?

qcelemental/models/procedures.py
```diff
@@ -10,14 +12,33 @@
...
+    final = "final"
+    none = "none"
+
+
+class OptimizationProtocol(ProtoModel):
+    """
+    Protocols regarding the manipulation of a Optimization output data.
+    """
+
+    keep_trajectory: KeepTrajectoryProtocolEnum = Schema(KeepTrajectoryProtocolEnum.all,
```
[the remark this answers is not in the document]
dgasmith: This relates back to the question if we want one protocol per field or not.

qcelemental/tests/test_models_results.py
dgasmith: `orbitals_a` is a pointer to the correct field. Non prefixed quantities are the expected return result and point to the actual field to save one space.

qcelemental/tests/test_models_results.py
dgasmith: We explicitly set lhs to be an index integer for this test.

mattwelborn:
1. What if protocols are simple -- a protocol transforms a small subset of the output in a simple way -- but multiple protocols may be provided as a list. The protocols would be applied in the order they appear in the list, and the user would be responsible for any interactions that caused. 
2. I lean towards no warnings. This way the program can produce as much output as it wants, and the protocol will trim it down. We won't need to write protocol handling into the program/harness directly. That said, there should probably be an error if a protocol asks for something and the program fails to produce it.

qcelemental/models/results.py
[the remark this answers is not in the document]
loriab: Just suggesting an alternate field name that can't conflict with python `all`

qcelemental/models/procedures.py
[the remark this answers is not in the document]
loriab: Perhaps I'm confused. I supposed that a QCSchemus would have one protocol field, which would be a dictionary and example contents would be `trajectory`: `first_and_last` and `wavefunction`: `orbitals_only` (didn't check labels). Psi (for instance) _could_ do the wfn truncation in the course of the gradient but isn't obliged to. Geometric _could_ do the trajectory truncation before returning to driver but isn't obliged to. Engine _will_ apply those two truncations if not already in place. So long as you're not trying to prod geometric/psi4 to do the protocol data layout changes themselves, I'd class the engine notice as logging info or debug, not warning.

qcelemental/models/results.py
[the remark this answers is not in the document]
dgasmith: I don't think this conflicts with all since it will always either be `Enum.all` or "all".

dgasmith:
@loriab @mattwelborn Ready for a deep review.

qcelemental/models/basis.py
```diff
@@ -52,6 +60,21 @@ def nfunctions(self) -> int:
...
+        """
+        Checks if the shell represents a contracted Gaussian or not.
+
+        Returns
+        -------
+        bool
+            True if the shell is contracted.
+        """
+
+        if (len(self.coefficients) != 1) and (len(self.angular_momentum) == 1):
```
mattwelborn: `return (len(self.coefficients) != 1) and (len(self.angular_momentum) == 1)`?

qcelemental/models/procedures.py
```diff
@@ -60,3 +82,27 @@ class Optimization(OptimizationInput):
...
+        keep_enum = values['protocols'].trajectory
+        if keep_enum == 'all':
+            pass
+        elif keep_enum == "initial_and_final":
+            if len(v) != 2:
+                v = [v[0], v[-1]]
+        elif keep_enum == "final":
+            if len(v) != 1:
+                v = [v[-1]]
+        elif keep_enum == "none":
```
mattwelborn: should this and above also set "initial_molecule" and "final_molecule" to None as appropriate? Or does that happen later?

qcelemental/models/results.py
```diff
@@ -151,10 +156,7 @@ def _assert1d(cls, v, values):
             raise ValueError("Vector must be castable to shape (-1, )!")
         return v
 
-    @validator(
-        'scf_orbitals_a',
-        'scf_orbitals_b',
-        whole=True)
+    @validator('scf_orbitals_a', 'scf_orbitals_b', whole=True)
     def _assert2d_nao_x(cls, v, values):
```
mattwelborn: scf_orbitals_b could be empty if restricted = True

qcelemental/models/results.py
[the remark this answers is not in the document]
mattwelborn: I vote for "all".

qcelemental/models/procedures.py
dgasmith: Modifying these would be different fields. `protocols: initial_molecule`.

qcelemental/models/results.py
dgasmith: Sure, this is accounted for since it happens *after* initial validation.

qcelemental/models/basis.py
dgasmith: Whoops :)

qcelemental/models/procedures.py
```diff
@@ -60,3 +82,27 @@ class Optimization(OptimizationInput):
...
+            if len(v) != []:
```
loriab: `!= 0`

just tried it:
```
>>> len([]) == []
False
```

dgasmith: No need for the check even, fixed.

PR #142, opened 2019-09-30
Update change log for 0.9.0 and add GitHub Templates

Lnaden:
As it says on the tin!


dgasmith:
LGTM, please cut a release if you have time. Else I will get to it tomorrow.

PR #143, opened 2019-10-07
Allow testing of complex number

### LINKED ISSUE #77: Testing complex numbers
I am using the testing functionality within Psi4 and I am running into this warning when using `compare_arrays`:
```
 /home/roberto/Workspace/robertodr/psi4/build_xtensor/stage/lib/qcelemental/testing.py:106: ComplexWarning: Casting complex values to real discards the imaginary part
  xptd, cptd = np.array(expected, dtype=np.float), np.array(computed, dtype=np.float)
```
for complex numbers this is unwanted, as I'd like to test that both real and imaginary parts are as expected.

I changed that line to:
```
xptd, cptd = np.array(expected, dtype=expected.dtype), np.array(computed, dtype=computed.dtype)
```
though it doesn't seem exactly correct to do this. Probably a check that the types match between expected and computed should be done before checking that the actual values match.

@loriab suggested getting rid of this line https://github.com/MolSSI/QCElemental/blob/master/qcelemental/datum.py#L47 but the warning still pops up.

robertodr:
Fixes #77 where warnings for narrowing conversions of complex numbers were being emitted.

## Todos
Notable points that this PR has either accomplished or will accomplish.
  - [x] Fix #77

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcelemental/testing.py
+++ b/qcelemental/testing.py
@@ -102,7 +102,7 @@ def compare_values(expected,
         if expected is None and computed is None:
             return return_handler(True, label, pass_message, return_message, quiet)
 
-    if expected.dtype == np.complex:
+    if np.iscomplexobj(expected):
         dtype = np.complex
     else:
         dtype = np.float
--- a/qcelemental/testing.py
+++ b/qcelemental/testing.py
@@ -102,10 +102,15 @@ def compare_values(expected,
         if expected is None and computed is None:
             return return_handler(True, label, pass_message, return_message, quiet)
 
+    if expected.dtype == np.complex:
+        dtype = np.complex
+    else:
+        dtype = np.float
+
     try:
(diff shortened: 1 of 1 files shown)


loriab:
I approved too soon -- travis failures are legit. Could you add a complex val and complex array to the testing testing, please?

qcelemental/testing.py
```diff
@@ -102,10 +102,15 @@ def compare_values(expected,
         if expected is None and computed is None:
             return return_handler(True, label, pass_message, return_message, quiet)
 
+    if expected.dtype == np.complex:
```
dgasmith: We need `np.iscomplexobj(expected)` here. Otherwise `complex64` etc is incorrectly typed.

robertodr: Good to know there's such a function in NumPy! Let's see what Travis has to say now.

PR #144, opened 2019-10-07
`which_import` now handles submodules correctly

dotsdl:
## Description

<strike>For checking if OpenMM exists in the environment, we need to feed
"simtk.openmm" to `importlib.util.find_spec`. This will work by actually
importing `simtk`, which in the case of OpenMM not being present raises
a `ModuleNotFoundError`.

This PR handles the case in which this `ModuleNotFoundError` is raised.</strike>

For checking if OpenMM exists in the environment, we need to feed "openmm" with `package="simtk"` to `importlib.util.find_spec`. This PR adds a new keyword arg `package` to `which_import` in order to pass this through to `find_spec`.

## Status
- [ ] Changelog updated
- [ ] Ready to go

### DIFF
--- a/qcelemental/util/importing.py
+++ b/qcelemental/util/importing.py
@@ -5,7 +5,7 @@ from typing import Union
 
 
 def which_import(module: str, *, return_bool: bool = False, raise_error: bool = False,
-                 raise_msg: str = None) -> Union[bool, None, str]:
+                 raise_msg: str = None, package: str = None) -> Union[bool, None, str]:
     """Tests to see if a Python module is available.
 
     Returns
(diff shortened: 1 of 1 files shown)


dotsdl:
This is currently blocking MolSSI/QCEngine#151.

loriab:
Hadn't thought about this case. Reading through [`find_spec`](https://docs.python.org/3/library/importlib.html#importlib.util.find_spec), would it be cleaner if we passed `package` arg through, so you could call `which_import("openmm", package=simtk)`?

dotsdl:
@loriab good catch! I'll give that a try.

dotsdl:
@loriab this appears to have the desired effect. Thanks!

PR #147, opened 2019-10-18
Pydantic 1.0 Changes

dgasmith:
## Description
Updates QCElemental to be Pydantic 1.0 compliant. 

@loriab There also seems to be a networkx release change, see 8dc5abc.

This should not be merged until Pydantic 1.0 is released. Our Conda package is pinned to <1.0 so we should be fine.

## Status
- [ ] Changelog updated
- [x] Ready to go


qcelemental/util/gph_uno_bipartite.py
```diff
@@ -198,10 +198,10 @@ def _enumMaximumMatchingIter(g, match, all_matches, add_e=None):
         new_match = []
         for ee in d.edges():
             if ee in len2path:
-                if g.node[ee[1]]['bipartite'] == 0:
+                if g.nodes[ee[1]]['bipartite'] == 0:
```
loriab: thanks, looks like a 2.4 change https://networkx.github.io/documentation/stable/release/release_2.4.html

PR #150, opened 2019-10-24
`which_import` now *really* handles submodules correctly

dotsdl:
## Description
This is a correction on #144.
This correction also includes tests to ensure the behavior works as originally intended.

Apologies for the churn; I should have built some explicit tests in the first go-around. My bleary-eyed *ad hoc* tests were not a good substitute.

## Status
- [ ] Changelog updated
- [ ] Ready to go

### DIFF
--- a/qcelemental/tests/test_importing.py
+++ b/qcelemental/tests/test_importing.py
@@ -39,6 +39,50 @@ def test_which_import_f_raisemsg():
     assert str(e.value).endswith("Python module 'evilpint' not found in envvar PYTHONPATH. Install `evilpint`.")
 
 
+def test_which_import_t_submodule():
+    ans = qcel.util.which_import('pint.util')
+    assert ans.split(os.path.sep)[-1] == 'util.py'
+
+
+def test_which_import_t_submodule_altsyntax():
+    ans = qcel.util.which_import('.util', package='pint')
+    assert ans.split(os.path.sep)[-1] == 'util.py'
+
+
+def test_which_import_t_bool_submodule():
+    ans = qcel.util.which_import('pint.util', return_bool=True)
+    assert ans is True
+
+
+def test_which_import_f_submodule():
+    ans = qcel.util.which_import('evilpint.util')
+    assert ans is None
+
+
+def test_which_import_f_submodule_altsyntax():
+    ans = qcel.util.which_import('.util', package='evilpint')
+    assert ans is None
+
+
+def test_which_import_f_bool_submodule():
+    ans = qcel.util.which_import('evilpint.util', return_bool=True)
+    assert ans is False
+
+
+def test_which_import_f_raise_submodule():
+    with pytest.raises(ModuleNotFoundError) as e:
+        qcel.util.which_import('evilpint.util', raise_error=True)
+
+    assert str(e.value).endswith("Python module 'evilpint.util' not found in envvar PYTHONPATH.")
+
+
+def test_which_import_f_raisemsg_submodule():
+    with pytest.raises(ModuleNotFoundError) as e:
+        qcel.util.which_import('evilpint.util', raise_error=True, raise_msg='Install `evilpint`.')
+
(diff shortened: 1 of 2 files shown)


qcelemental/tests/test_importing.py
```diff
@@ -39,6 +39,50 @@ def test_which_import_f_raisemsg():
...
 
 
+def test_which_import_t_submodule():
+    ans = qcel.util.which_import('pint.util')
+    assert ans.split(os.path.sep)[-1] == 'util.py'
+
+
+def test_which_import_t_submodule_altsyntax():
+    ans = qcel.util.which_import('.util', package='pint')
+    assert ans.split(os.path.sep)[-1] == 'util.py'
```
loriab: I would have thought that only this case (of the three positive) ones was needed. If we want to permit `which_import('pint.util')`, can take the arg and `module.rsplit('.', 1)` it and send the `[0]` to `package=`.

qcelemental/tests/test_importing.py
```diff
@@ -39,6 +39,50 @@ def test_which_import_f_raisemsg():
...
+    assert ans is True
+
+
+def test_which_import_f_submodule():
+    ans = qcel.util.which_import('evilpint.util')
+    assert ans is None
+
+
+def test_which_import_f_submodule_altsyntax():
+    ans = qcel.util.which_import('.util', package='evilpint')
```
loriab: to test something new, perhaps you want `which_import('.evilutil', package='pint')`

qcelemental/util/importing.py
```diff
@@ -23,7 +23,10 @@ def which_import(module: str, *, return_bool: bool = False, raise_error: bool =
     """
     import importlib
 
-    module_spec = importlib.util.find_spec(module, package=package)
+    try:
+        module_spec = importlib.util.find_spec(module, package=package)
+    except ModuleNotFoundError:
+        module_spec = None
```
loriab: (from above) I think the try/catch can be avoided

dgasmith:
@loriab Any real blockers here that cannot be cleaned up in another version or patch?

loriab:
No, go for it.

qcelemental/util/importing.py
dotsdl: It turned out that the following uses of `importlib.util.find_spec` produce the same result:
```python
import importlib

importlib.util.find_spec("pint.util")

importlib.util.find_spec(".util", package="pint")
```
Both will produce a `ModuleNotFoundError`, which must be caught to handle for cases like `simtk.openmm`. I didn't properly notice this in #144 as doing

```
importlib.util.find_spec("util", package="pint")
```
does not throw an exception, but it also doesn't find the package, either (because we need the relative `'.util'`, not just `'util'`). It didn't register in my mind at the time that this didn't solve our problem.

The issue is that any attempt to use `find_spec` for a submodule always imports the module above first, and if it isn't present, we will get an exception. So, we have to catch it.

qcelemental/tests/test_importing.py
dotsdl: We could add this, yes. It would be testing a very different case, however, that won't yield a `ModuleNotFoundError`, since `pint` would exist, even though the `.evilutil` submodule doesn't.

qcelemental/tests/test_importing.py
dotsdl: It turns out that sending to the `package` kwarg actually doesn't change 'importlib.util.find_spec`'s behavior as we (I) had thought. Apologies for the mistake on my part. It turns out that there really isn't a way around a try-catch here (see my comment below).

qcelemental/util/importing.py
loriab: Thanks, I think I understand better now. Ok with the try/except. Still think a `pint.evilutil` test would be worthwhile. And you may need to adapt the default msg https://github.com/MolSSI/QCElemental/pull/150/files#diff-8e6fced7103e77b67c18f7cd16df2298R34 for when both module and package are passed so that user not advised to install a non-top-level pkg.

PR #151, opened 2019-10-24
v11 Changelog and Docs

dgasmith:
## Description
Mostly a changelog, but a light docs pass as well.

## Status
- [x] Changelog updated
- [x] Ready to go

### DIFF
--- a/docs/source/model_result.rst
+++ b/docs/source/model_result.rst
@@ -1,0 +1,27 @@
+Result
+======
+
+A Python implementation of the `MolSSI QCSchema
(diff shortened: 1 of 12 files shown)


docs/source/model_result.rst
```diff
@@ -0,0 +1,27 @@
+Result
```
loriab: how soon is this shifting to `SingleResult`? worth doing `Result (SingleResult)` and then reversing after the shift?

dgasmith: Let's put the name up for a vote then switch.

PR #152, opened 2019-10-25
Molecule: read and write xyz+, implement kwargs for from_data/file

mattwelborn:
## Description
This PR fixes xyz+ reading and writing for Molecules and makes kwargs work in from_data and from_file. This is motivated by difficulties encountered when specifying charge and spin states.

## Status
- [ ] Changelog updated
- [ ] Ready to go


qcelemental/models/molecule.py
```diff
@@ -712,7 +713,7 @@ def from_data(cls,
                 raise TypeError("Input type not understood, please supply the 'dtype' kwarg.")
 
         if dtype in ["string", "psi4", "psi4+", "xyz", "xyz+"]:
-            mol_dict = from_string(data)
+            mol_dict = from_string(data, dtype if dtype is not "string" else None)
```
loriab: `!=`, not `is not`, I think.

qcelemental/models/molecule.py
```diff
@@ -773,7 +780,7 @@ def from_file(cls, filename: str, dtype: Optional[str] = None, *, orient: bool =
                 dtype = "string"
 
         # Raw string type, read and pass through
-        if dtype in ["string", "xyz", "psi4"]:
+        if dtype in ["string", "xyz", "xyz+", "psi4"]:
```
loriab: now you point it out, should probably drop `psi4+` from https://github.com/MolSSI/QCElemental/pull/152/files#diff-c1df0b88317a0c0331bea59a1793059cR715, too. it will only collect a spec that `model.Molecule` can't handl.e

qcelemental/molparse/to_string.py
```diff
@@ -127,7 +128,11 @@ def to_dict(self) -> Dict:
         nat = len(atoms)
 
         first_line = """{} {}""".format(str(nat), umap.get(units.lower(), units.lower()))
-        smol = [first_line.rstrip(), name]
+        smol = [first_line.rstrip()]
+        if dtype == 'xyz':
+            smol.append(name)
+        else:
+            smol.append(f"{molrec['molecular_charge']} {molrec['molecular_multiplicity']} {name}")
```
loriab: chg mult name isn't a violation of "xyz" format. and "xyz" is already violated by units in the first line, so there's nothing but a choice stopping you from writing the extended seconda line for both xyz and xyz+. up to you.

qcelemental/models/molecule.py
```diff
@@ -738,6 +739,12 @@ def from_data(cls,
         else:
             raise KeyError("Dtype not understood '{}'.".format(dtype))
 
+        input_dict.update(kwargs)
+        if "molecular_charge" in kwargs and "fragment_charges" not in kwargs:
+            input_dict["fragment_charges"] = None
+        if "molecular_multiplicity" in kwargs and "fragment_multiplicities" not in kwargs:
+            input_dict["fragment_multiplicities"] = None
```
loriab: probably if any _one_ of the four is present in kwargs, all need to be `None`d, since chgmult not independent.

qcelemental/molparse/to_string.py
mattwelborn: Agreed.

mattwelborn: In order to make this change, I had to update some tests. Could you please take a look and make sure the new test reference results are correct? I don't know all of the atom symbol annotations used.

mattwelborn:
@loriab Thanks for the review!

dgasmith:
@loriab Please re-review and merge when you are ready.

PR #156, opened 2019-11-01
Native Molecule Hashing

### LINKED ISSUE #154: Native Molecule Hashing
`mol1 == mol2` should use `get_hash()` not whatever comparison is native to Pydantic.

Override `__hash__` in Molecule to call `get_hash()` and maybe deprecate that function so that `hash(Molecule)` works just as well.

@mattwelborn

mattwelborn:
## Description
This PR implements `__eq__` for Molecule by comparing hashes from `get_hash`. `get_hash` was chosen over `__hash__` because our Molecule hashes do not satisfy Python's requirements. `Molecule.compare` is deprecated. Resolves #154.

## Status
- [x] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcelemental/models/molecule.py
+++ b/qcelemental/models/molecule.py
@@ -5,6 +5,7 @@ Molecule Object Model
 import collections
 import hashlib
 import json
+import warnings
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, Union
 
@@ -388,6 +389,11 @@ class Molecule(ProtoModel):
         return Molecule(orient=True, **self.dict())
 
     def compare(self, other):
+        warnings.warn("Molecule.compare is deprecated and will be removed in v0.13.0. Use == instead.",
+                      DeprecationWarning)
+        return self == other
+
+    def __eq__(self, other):
         """
         Checks if two molecules are identical. This is a molecular identity defined
         by scientific terms, and not programing terms, so it's less rigorous than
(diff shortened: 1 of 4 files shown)


qcelemental/models/molecule.py
```diff
@@ -401,18 +407,7 @@ def compare(self, other):
...
-        match &= np.array_equal(self.real, other.real)
-        match &= np.array_equal(self.fragments, other.fragments)
-        match &= np.allclose(self.fragment_charges, other.fragment_charges, atol=(10**-CHARGE_NOISE))
-        match &= np.array_equal(self.fragment_multiplicities, other.fragment_multiplicities)
-
-        match &= np.allclose(self.molecular_charge, other.molecular_charge, atol=(10**-CHARGE_NOISE))
-        match &= np.array_equal(self.molecular_multiplicity, other.molecular_multiplicity)
-        match &= np.allclose(self.geometry, other.geometry, atol=(10**-GEOMETRY_NOISE))
-        return match
+        return self.get_hash() == other.get_hash()
```
dgasmith: Do we want to allow `compare` to stay in to allow tunable tolerances? I can think of a few use cases, but then again we have `compare_recursive` which can handle this.

@loriab any thoughts here?

qcelemental/models/molecule.py
```diff
@@ -388,6 +389,11 @@ def orient_molecule(self):
         return Molecule(orient=True, **self.dict())
 
     def compare(self, other):
+        warnings.warn("Molecule.compare is deprecated and will be removed in v0.13.0. Use == instead.",
+                      DeprecationWarning)
```
loriab: fwiw, this is the form we settled on for p4. `FutureWarning` https://github.com/psi4/psi4/blob/02bc2cbc17e8779b1cd6b0aa687760cee59e6fa8/psi4/driver/p4util/python_helpers.py#L711-L714

qcelemental/models/molecule.py
loriab: I can think of lots o'ways ppl want to judge mol equality (connectivity-wise, geometry-wise, geometry-w/i-tol-wise, geometry-ignoring-frame-and-orientation-wise, equal-except-isotope, equal-except-chgmult, equal-ignoring-fragmentation, equal-ignoring-ghostiness, equal-except-different-bases-on-atoms, equal-but-for-labels, equal-but-for-atom-order). Looks like existing `compare` is focusing on just variations in numerics. I don't think I use `models.Molecule.compare`, so no harm to my downstream in dropping it.

Someday, may want to coalesce some of the above into a tunable fn (with some presets like "is_isotope"), but I'm fine with dropping for now to focus on the broadly used preset hash. Being a built-in, that'll not be tunable, so a separate fn will make sense someday.

PR #162, opened 2019-11-05
add ResultProtocols.logfile

loriab:
## Todos
  - [x] following slack discussion, use a protocol to control stdout return rather than `return_output` field (which doesn't validate with no extras)
  - [x] move `DriverEnum` up to `models`. I import it a lot, and most else is common_models was accessible

## Questions
- [ ] This could be made an enum and control stdout/stderr/other returned files. But keeping it simple for now.

## Status
- [x] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcelemental/models/results.py
+++ b/qcelemental/models/results.py
@@ -220,6 +220,7 @@ class ResultProtocols(ProtoModel):
 
     wavefunction: WavefunctionProtocolEnum = Field(WavefunctionProtocolEnum.none,
                                                    description=str(WavefunctionProtocolEnum.__doc__))
+    logfile: bool = Field(True, description="Primary output file to keep from a Result computation")
 
     class Config:
         force_skip_defaults = True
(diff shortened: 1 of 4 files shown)


qcelemental/models/results.py
```diff
@@ -220,6 +220,7 @@ class ResultProtocols(ProtoModel):
 
     wavefunction: WavefunctionProtocolEnum = Field(WavefunctionProtocolEnum.none,
                                                    description=str(WavefunctionProtocolEnum.__doc__))
+    logfile: bool = Field(True, description="Primary output file to keep from a Result computation")
```
dgasmith: This needs to be called `stdout` to match the name of the field.

qcelemental/models/results.py
```diff
@@ -347,3 +348,18 @@ def _wavefunction_protocol(cls, value, values):
             return ret_wfn
         else:
             return wfn
+
+    @validator('stdout', pre=True)
+    def _logfile_protocol(cls, value, values):
```
dgasmith: Does this one have to be a `pre`? Sorry, I forget exactly why the others needed to be.

loriab: I didn’t think so when I looked up pre but erred on the side of matching other protocol. Will try w/o.

Issue #155, opened 2019-10-30
Result --> SingleResult


loriab:
So we've got
* `OptimizationInput`/`Optimization`
* `ResultInput`/`Result`

And in another project, I've got the potential for
* `FindifResultInput`/`FindifResult`
* `CBSResultInput`/`CBSResult`

etc. The optimization pair doesn't particularly signal "schema" to me. And it might be convenient to distinguish between the set of `*ResultInput`/`*Result` pairs and the particular energy/gradient/hessian calc that `ResultInput`/`Result` now represents.

It is suggested that the two first items above become the below so that the `*` of `*ResultInput`/`*Result` is never the empty string.
* `OptimizationResultInput`/`OptimizationResult`
* `SingleResultInput`/`SingleResult` (`AnalyticResult`?)

Please weigh in on this change, even if it's just an up/down emoji response.

mattwelborn:
@dgasmith Does this proposal interact at all with plans to make optimization a service?

dgasmith:
@mattwelborn Nope, no changes from me.

Another option is to do:
 - `OptimizationInput / Optimization`
 - `SingleResultInput / SingleResult`
 - ...

Since these names are going to be typed many times I would be hesitant to make them too long. 
`FindifInput/Findif`, `CBSInput/CBS`

jchodera:
I'm with @dgasmith here: It's clearer to use `XInput/XResult`, or perhaps even `XInput/XOutput`. `ResultInput` is syntactically painful.

loriab:
I've thought `ResultInput` a pleasing quirk because you could search for `CBSResult` and hit both input and output stages. But sure, I don't mind `XInput/XResult`. It's searchable and predictable (latter not the case at present). 

So any votes among the below? I'm plenty happy with SingleResult. It's distinctive but with a vague enough definition to not feel wrong stretching it.
* `SingleResult`
* `AnalyticResult`
* `PrimaryResult`
* `AtomicResult`

dgasmith:
`QuantumResult` for the word play, both *a* quantum unit and a quantum chemistry one. I will see myself out...

dgasmith:
`QuantaResult`?

loriab:
[...]

PR #167, opened 2019-11-08
Models: Consistent model nomenclature moving forward

dgasmith:
## Description
Renames models according to #155.

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcelemental/models/results.py
+++ b/qcelemental/models/results.py
@@ -12,7 +12,7 @@ from .molecule import Molecule
 from .types import Array
 
 
-class ResultProperties(ProtoModel):
+class AtomicResultProperties(ProtoModel):
     """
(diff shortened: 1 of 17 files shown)


qcelemental/models/results.py
```diff
@@ -424,3 +425,33 @@ def _stdout_protocol(cls, value, values):
             return None
         else:
             raise ValueError(f"Protocol `stdout:{outp}` is not understood")
+
+
+def ResultProtocols(*args, **kwargs):
+    from warnings import warn
+
+    warn("ResultInput has been renamed to AtomicResultProtocols and will be removed in v0.13.0", DeprecationWarning)
```
loriab: "ResultProtocols has been renamed ..."

PR #177, opened 2019-12-12
Missing imports and a few typos

sjrl:
## Description
I noticed some missing imports and a few typos so I made a few changes. Please let me know if this helpful!

## Status

- [x] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcelemental/molutil/align.py
+++ b/qcelemental/molutil/align.py
@@ -263,6 +263,7 @@ def B787(
             print(atomfmt2.format(auniq[at][:6], *[ageom[at][i] - rgeom[at][i] for i in range(3)]))
 
     if do_plot:
+        # TODO Missing import
         plot_coord(ref=rgeom, cand=ageom, orig=cgeom, comment="Final RMSD = {:8.4f}".format(final_rmsd))
 
     # sanity checks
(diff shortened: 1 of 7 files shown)


qcelemental/molutil/align.py
```diff
@@ -263,6 +263,7 @@ def _plausible_atom_orderings_wrapper(
             print(atomfmt2.format(auniq[at][:6], *[ageom[at][i] - rgeom[at][i] for i in range(3)]))
 
     if do_plot:
+        # TODO Missing import
```
loriab: yeah, this is a relic from when I was developing the code in jupyter and had the target, before, and after mols pop up for sanity checking. possibly this should be hidden further now.

loriab:
CI fails look to be pubchem timeouts. They happen.

dgasmith:
`TYPE_CHECKING` is always hand written and should be validated by `mypy`. Always curious if we should enforce `mypy` checking which will add quite a bit of complexity.

loriab:
gtk, I thought some `TYPE_CHECKING` blocks materialized when I ran `make format`. I don't feel strongly on `mypy`. If it was easy to enforce by tool, I'd be willing. But I'm also satisfied by semiannual `mypy` passes.

PR #179, opened 2020-01-06
pint 0.10, python 3.8, year 2020

loriab:
## Description
pint 0.10, python 3.8, year 2020

## Changelog description
Already updated

## Status

- [x] Code base linted
- [x] Ready to go


.github/workflows/CI.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         conda-env: [base, minimal]
-        python-version: [3.6, 3.7]
+        python-version: [3.6, 3.7, 3.8]
```
dgasmith: Can we just run 3.6/3.8, will keep the total number of builds down.

loriab:
CI-wise, I think this is ready to merge. There's not anything preventing a conda env with qcel 0.12 (last release) and py38, so I don't see that a version bump is needed immediately or that the existing c-f package needs rebuilding, though the pint constraint should be added to the recipe at next edit.

PR #181, opened 2020-01-07
numpy 1.12

loriab:
## Description
mention numpy>=1.12 constraint inherited from pint. distribution-wise, c-f is home free b/c it builds against a minimum 1.14. but default's minimum is 1.11, so worth having explicit constraint. 

## Changelog description
already edited

## Status

- [x] Code base linted
- [x] Ready to go (if CI passes)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # QCElemental
 
 [![Build Status](https://github.com/MolSSI/QCElemental/workflows/CI/badge.svg?branch=master)](https://github.com/MolSSI/QCElemental/actions?query=workflow%3ACI)
-[![codecov](https://codecov.io/gh/MolSSI/QCElemental/branch/master/graph/badge.svg)](https://codecov.io/gh/MolSSI/QCElemental)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1,7 +1,7 @@
 # QCElemental
 
 [![Build Status](https://github.com/MolSSI/QCElemental/workflows/CI/badge.svg?branch=master)](https://github.com/MolSSI/QCElemental/actions?query=workflow%3ACI)
-[![codecov](https://codecov.io/gh/MolSSI/QCElemental/branch/master/graph/badge.svg)](https://codecov.io/gh/MolSSI/QCElemental)
+[![codecov](https://img.shields.io/codecov/c/github/MolSSI/QCElemental.svg?logo=Codecov&logoColor=white)](https://codecov.io/gh/MolSSI/QCElemental)
```
dgasmith: Why did we change this to a non-native badge?

loriab: I haven’t favored native over shieldsio badges, and the latter are more customizable. This fixes the hard to read dark pink on black umbrella (matches psi).

dgasmith: Ok, would mind propagating this to the rest of the ecosystem?

PR #182, opened 2020-01-07
Added protocol for controlling error retries

WardLT:
## Description
Adds a new protocol for controlling the error correction feature added in molssi/qcengine#210.

## Changelog description
Add protocol for controlling error autocorrection

## Status

- [x] Code base linted
- [ ] Ready to go


qcelemental/models/results.py
```diff
@@ -270,6 +270,12 @@ class AtomicResultProtocols(ProtoModel):
         WavefunctionProtocolEnum.none, description=str(WavefunctionProtocolEnum.__doc__)
     )
     stdout: bool = Field(True, description="Primary output file to keep from a Result computation")
+    error_correction_retries: int = Field(
```
dgasmith: Unsure if this particularly a count issue. Imagine I need to fix three issues similar to "autosym" in NWChem, this feels more like a "fix all inconsequential errors" enum. "inconsequential" of course will be something hard to define.

This is separate from try to fix convergence through 1-3 ranks:
 - rank 1, default
 - rank 2, SOSCF
 - rank 3, SOSCF + GDIIS + level shift

loriab: Second the "count" comment, as I was typing that ppl would likely have this at 0 or 50 because correctable errors of this tweak type are independent. So perhaps a bool would do? The 1-3 ranks Daniel describes are perhaps a `GuidableError` that are managed by separate protocols.

WardLT: To make sure I follow your suggestion correctly, is the idea to have the user specify a "level of consequentialness." Where higher levels will allow QCEngine to make more aggressive changes to an input file?

dgasmith: I think so. I say *think* here because I am still working through the consequences of this. At the moment, I am working through custodian pretty carefully and pinging the MP team with questions.

loriab: My notion was that all the inconsequential `CorrectableError`s would be controlled by a single bool. these would be the type that an experienced practitioner walks up to the troublesome input as says, "ah, add autosym there", and walks away unconcerned about the consequences. the more involved "try this, study the output, try that, check the output for red flag" would each be separately controlled. Daniel's rank 1-3 example is in this category.

WardLT:
Oh no, I rebased with MOLSSI's master and now the commit history looks insane. Do you want me to fix this?

loriab:
Ah, you needed to rebase a force push but you probably rebased pulled and pushed?

WardLT:
Yea, I already rebase pull and pushed. Sorry, I'm still getting better with git. :\

Any suggestions for resolving, or should I leave this as-is?

qcelemental/checkup_data/physconst.py
```diff
@@ -31,5 +31,5 @@
 hartree2MHz               = 6.579684E9           # Hartree to MHz conversion factor       
 kcalmol2wavenumbers       = 349.7551             # kcal mol$^{-1}$ to cm$^{-1}$ conversion factor 
 e0                        = 8.854187817E-12      # Vacuum permittivity (Fm$^{-1}$)        
-na                        = 6.02214179E23        # Avagadro's number                      
+na                        = 6.02214179E23        # Avagadro's protocol
```
dgasmith: I think this sed might have gone a bit too wide :)

qcelemental/models/results.py
```diff
@@ -270,6 +281,14 @@ class AtomicResultProtocols(ProtoModel):
         WavefunctionProtocolEnum.none, description=str(WavefunctionProtocolEnum.__doc__)
     )
     stdout: bool = Field(True, description="Primary output file to keep from a Result computation")
+    error_correction_policy: ErrorCorrectionEnum = Field(
+        ErrorCorrectionEnum.all, description="Policy used to determine which error correction routines to use"
+    )
+    error_correction_settings: Optional[Dict[str, bool]] = Field(
```
dgasmith: This looks like a whitelist/blacklist and a True/False per setting which would be similar to the white/blacklisting? Perhaps `ErrorCorrectionEnum.explicit` for this or am I way off base.

WardLT: Yea, this just occurred to me while I as writing a convoluted description of the changes I made. I'm working on a simplified version

WardLT:
Alright, this is ready for review now.

You specify error policies with two settings in the AtomicResultProtocol, 

- `error_correction_policy` defines whether specific error corrections are allowed to be used
- `error_correction_default_policy` defines what you should do in the absence of specific instruction

qcelemental/models/results.py
loriab: Maybe I'm planning UI too far in advance, but for the enum, I was thinking none/all/safe/.../explicit so we can group common ones like "orbitals_and_eigenvalues" does for wfn.

WardLT: Would the idea be that we can label specific error corrections with tags such as "safe"?

loriab: I wasn't thinking labels on each correction as just shortcuts to groups https://github.com/MolSSI/QCElemental/blob/master/qcelemental/models/results.py#L395-L396. But the dict is the more important at this stage

WardLT: I see. That sort of a single level identifier mapping to more than one type of error could be a good model and would be the most beneficial if we have many engines that throw the same type of error. 

I was picturing attaching some kind of "danger level" to certain error corrections, depending on how likely they are to affect the accuracy of the output. That would be more beneficial if we have different error correction routines each code. 

I'm not sure which of these pictures (if accurate) is more accurate. Should we punt on this until we have more error classes to go off of?

loriab: I think both the pictures you describe will likely show up (the symm cmd that started all this is nwc only but use-more-opt-cycles will be valid for many progs). I'm just advocating that the non-dict field not be a bool but instead be an enum (none/all/explicit for starters). But if you and @dgasmith are happy with bool for now, I won't insist.

WardLT:
Sorry for letting this PR drag.

I think we should keep the default policy as a boolean for now. I do like @loriab 's suggestion of allowing for a string designation of common policies, but would prefer to keep the simpler implementation as I yet have a need for defining common subsets of error correction policies. I put a note in the source code that we should consider changing this implementation in the future. 
That said, I would not mind being overruled.

Since the last review, I moved the logic for determining whether a specific policy is accepted into QCElemental. That seemed to make more sense than having it in QCEngine

WardLT:
@loriab / @dgasmith Any thoughts on whether this is good to merge? I've merged in the recent `master`

WardLT:
I'm not in a hurry to merge this and would prefer make sure we are both on board.

I think this is the comment where we seem out of agreement:
> I'm just advocating that the non-dict field not be a bool but instead be an enum (none/all/explicit for starters).

I've made a few drafts of adding `none/all/explicit`, but was not sure if I have the correct understanding. Under `none`, would you expect the ability to override the policy and allow certain error corrections? If so, how would this be different than `explicit`?

loriab:
I was thinking of 'none' as a reject-all-help-and-behave-like-raw-qc-software option, 'all' as get-me-a-result-if-you-can, and 'explicit' as whitelist-these.

But I'm also good with noting this as experimental and getting it into the wild.

WardLT:
My vote is to keep this as-is and we can revisit as more people use it. Shall I add in some language around this being expeirmental into the source code?

loriab:
Sounds good. We don't have a proper "experimental" decorator, so yes, docstring or elsewhere for experimental alerts.

PR #190, opened 2020-01-12
try minimal storage mol for atomic_number

loriab:
## Description

Start of minimal storage for mol. Minimal validation is definitely a target, but I understand minimal storage is first. I started passing up the defaulted fields up but then figured why bother passing when it can be computed as needed. This is a first stab at #169 dealing only with atomic_number. Autogeneration does work or the nuclear_repulsion_energy fn test would fail. Currently 12 tests fail, mostly with extra_fields/serialization trouble. Looking for first comments.

## Status

- [x] Code base linted
- [ ] Ready to go

PR #191, opened 2020-01-13
reduced mol storage 2

loriab:
## Description
Continuation of #190 with all fields targeted. This is strictly test driven dev on existing test suite. All pass but 3 and those are ok if turn off default filtering. The 3 that don't pass I don't understand b/c masses are perfectly accessible (https://github.com/MolSSI/QCElemental/compare/master...loriab:smollmol2?expand=1#diff-c1df0b88317a0c0331bea59a1793059cR769) but not getting stashed for hashing.

```
[ 1.00782503223 15.99491461957 15.99491461957  1.00782503223]
superdict
schema_name qcschema_molecule
schema_version 2
validated True
symbols ['H' 'O' 'O' 'H']
geometry [[ 1.7317  1.2909  1.0371]
 [ 1.3156 -0.0074 -0.2807]
 [-1.3143  0.0084 -0.2741]
 [-1.7241 -1.3079  1.0277]]
name H2O2
identifiers None
comment None
molecular_charge 0.0
molecular_multiplicity 1
masses None
real None
atom_labels None
atomic_numbers None
mass_numbers None
connectivity None
fragments None
fragment_charges None
fragment_multiplicities None
fix_com False
fix_orientation False
fix_symmetry None
provenance {'creator': 'QCElemental', 'version': 'v0.12.0+59.gce318ac', 'routine': 'qcelemental.molparse.from_schema'}
id None
extras None
symbols ['H' 'O' 'O' 'H']
masses None
```

## Status

- [x] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcelemental/models/molecule.py
+++ b/qcelemental/models/molecule.py
@@ -145,7 +145,7 @@ class Molecule(ProtoModel):
     molecular_multiplicity: int = Field(1, description="The total multiplicity of this Molecule.")  # type: ignore
 
     # Atom data
-    masses: Optional[Array[float]] = Field(  # type: ignore
+    masses_: Optional[Array[float]] = Field(  # type: ignore
         None,
         description="An ordered 1-D array-like object of atomic masses [u] of shape (nat,). Index order "
         "matches the 0-indexed indices of all other per-atom settings like ``symbols`` and ``real``. If "
@@ -153,7 +153,7 @@ class Molecule(ProtoModel):
         "is provided, it must be the same length as ``symbols`` but can accept ``None`` entries for "
         "standard masses to infer from the same index in the ``symbols`` field.",
     )
-    real: Optional[Array[bool]] = Field(  # type: ignore
+    real_: Optional[Array[bool]] = Field(  # type: ignore
         None,
         description="An ordered 1-D array-like object of shape (nat,) indicating if each atom is real (``True``) or "
         "ghost/virtual (``False``). Index "
@@ -181,13 +181,13 @@ class Molecule(ProtoModel):
     )
 
     # Fragment and connection data
-    connectivity: Optional[List[Tuple[int, int, float]]] = Field(  # type: ignore
+    connectivity_: Optional[List[Tuple[int, int, float]]] = Field(  # type: ignore
         None,
         description="The connectivity information between each atom in the ``symbols`` array. Each entry in this "
         "list is a Tuple of ``(atom_index_A, atom_index_B, bond_order)`` where the ``atom_index`` "
         "matches the 0-indexed indices of all other per-atom settings like ``symbols`` and ``real``.",
     )
-    fragments: Optional[List[Array[np.int32]]] = Field(  # type: ignore
+    fragments_: Optional[List[Array[np.int32]]] = Field(  # type: ignore
         None,
         description="An indication of which sets of atoms are fragments within the Molecule. This is a list of shape "
         "(nfr) of 1-D array-like objects of arbitrary length. Each entry in the list indicates a new "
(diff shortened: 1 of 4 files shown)


qcelemental/models/molecule.py
```diff
@@ -675,11 +766,16 @@ def get_hash(self):
...
-        tmp_dict = super().dict(exclude_unset=False)
+        print(self.masses)
+        tmp_dict = super().dict(exclude_unset=False, by_alias=True)
+        print("superdict")
+        for k, v in tmp_dict.items():
+            print(k, v)
 
         np.set_printoptions(precision=16)
         for field in self.hash_fields:
             data = tmp_dict[field]
```
dgasmith: Maybe try `getattr(self, field)`?

loriab:
now working test-wise. glad you're finding it simpler than expected. I'm finding it more involved. But then you've been thinking about this end, whereas I was worried about the backend, which I ended up punting on. I expect some more testing is in order.

It's a little out-of-pattern that the connectivity property returns None for the default. But I guess I'm coming around to that.

dgasmith:
I view it as `None` mean no connectivity supplied while `[]` means that there are no connected atoms (e.g., helium complexes).

loriab:
after many trails, tests working again. failure is pubchem, so no problem. most controversial changes of this round are `force_skip_defaults: bool = True` and dropping the need-fragments-to-specify-fragment_charges check (the "from_data_kwargs" test is why). It has a few are-we-getting-the-fields-we-want tests now, but more would probably be good. ready for a review round again.

qcelemental/models/molecule.py
```diff
@@ -811,7 +852,8 @@ def from_data(
         kwarg_keys = set(kwargs.keys())
         if len(charge_spin_opts & kwarg_keys) > 0:
             for key in charge_spin_opts - kwarg_keys:
-                input_dict[key] = None
+                input_dict.pop(key, None)
+            input_dict.pop("validated", None)
```
loriab: thinking of ways to lose data, one could `Molecule.from_data(str, masses=[something exotic])` and the masses wouldn't get considered in the __init__ validation now that we're no longer double molparse-ing here and in __init__.

qcelemental/models/molecule.py
```diff
@@ -1310,18 +1352,30 @@ def scramble(
 
 
 def _filter_defaults(dicary):
-    default_mass = np.array([periodictable.to_mass(z) for z in dicary["atomic_numbers"]])
+    #    print("\nfilter pree:", sorted(dicary))
```
loriab: reminder to chop

qcelemental/tests/test_molecule.py
```diff
@@ -617,3 +630,60 @@ def test_molecule_connectivity():
...
+    if extra_keys is not None:
+        expected_keys |= extra_keys
+
+    diff_keys = mol.dict().keys() ^ expected_keys
+    assert len(diff_keys) == 0, f"Diff Keys {diff_keys}"
+
+
+def test_sparse_molecule_connectivity():
+    """
+    A bit of a weird test, but because we set connectivity it should carry through.
```
loriab: you recognize that the same doesn't hold for https://github.com/MolSSI/QCElemental/pull/191/files#diff-2025ec5a9d08fc78a8e5239d9c715ef8L250-L251?

loriab:
Is it worth having a do-store-all argument? or is `mol_instance.dict(force_skip_defaults=False)` a thing? Or storing physconst and periodictable context names under which drop decisions were made?

qcelemental/tests/test_molecule.py
dgasmith: The link doesn't seem to work, where is this pointing to? But yes, I realize that connectivity is a special case. I wanted to pull it out for display so that we can modify in the future.

loriab: the latter two "fragment_" args in 
```def test_from_data_kwargs():
    mol = Molecule.from_data(
        """
        O 0 0 0
        H 0 1.5 0
        H 0 0 1.5
        """,
        molecular_charge=1,
        molecular_multiplicity=2,
        fragment_charges=[1],
        fragment_multiplicities=[2],
```

ok, good in connectivity as special case.

loriab:
> Not sure if a history is needed here. I think a rule of thumb is that if we need the history we are doing something wrong.

The thing I acknowledge I'm doing wrong is neglecting to record deep in molparse whether masses came from user or defaults, thus user == current defaults not getting stored. I'm ok with considering current PR a first pass and I'll ultimately supply that user-specified-masses-so-keep info to pydantic. The "history" was a stopgap to not losing info. Fine w/o it.

PR #197, opened 2020-01-19
Units: add atomic unit of time

mattwelborn:
## Description
This PR adds the atomic unit of time to the units registry as `au_time`. 

## Changelog description
The atomic unit of time has been added to the units registry (named `"au_time"`).

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcelemental/info/cpu_info.py
+++ b/qcelemental/info/cpu_info.py
@@ -5,8 +5,8 @@ Contains metadata about Processors
 import difflib
 import re
 from enum import Enum
-from typing import List, Optional
 from functools import lru_cache
+from typing import List, Optional
 
 from pydantic import Field
 
--- a/qcelemental/physical_constants/ureg.py
+++ b/qcelemental/physical_constants/ureg.py
@@ -43,6 +43,9 @@ def build_units_registry(context):
     ureg.define("wavenumber = 1 / centimeter")
     ureg.define("Angstrom = angstrom")
 
+    # Time
+    ureg.define(f"au_time = {phys_const['atomic unit of time']['value']} * {phys_const['atomic unit of time']['unit']}")
+
     # Masses
     ureg.define(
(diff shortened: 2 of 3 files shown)


mattwelborn:
@erikkjellgren @peter-reinholdt

mattwelborn:
@dgasmith What's going on inside physical_constants/context.py w.r.t. units in aliases? Specifically, why do some aliases not have units (e.g. `dipmom_au2debye`)? And why do some aliases appear to have units inconsistent with their description (e.g. `c` has units of Hz rather than m/s)?

mattwelborn:
@dgasmith Also, how would you feel about adding all of the atomic units with generic names (e.g. `au_distance`, `au_charge`, `au_energy`) just to make it easy on people so they don't have to remember Bohr, `e`, Hartree etc.?

peter-reinholdt:
Some of the useful derived units are
```
au_time: hbar/E_h
au_force: E_h  / a_0
au_current: e*E_h / hbar
au_electric dipole moment: e*a_0
au_magnetic dipole moment: hbar*e/m_e
```

erikkjellgren:
> @dgasmith Also, how would you feel about adding all of the atomic units with generic names (e.g. `au_distance`, `au_charge`, `au_energy`) just to make it easy on people so they don't have to remember Bohr, `e`, Hartree etc.?

I think this would be very nice. 
Also because you do not have to worry about other special names for atomic unit dimensions.

"hbar" could also be a nice constant to have.

dgasmith:
I think the `au_*` aliases are fine if it's more convenient, one note is that all words need to be connected with `_` so `au_magnetic dipole moment` -> `au_magnetic_dipole_moment`.

@loriab might have more comments on aliases, but effectively they were selected as an exact mapping to a known NIST unit as you need to define all aliases in one pass (`c` cannot be defined after the fact). This seems like a bit of a hack, but is totally valid. Cannot tell you on `dipmom_au2debye`.

dgasmith:
@mattwelborn Merge when ready, we can split the conversation into an issue or Slack.

PR #199, opened 2020-01-20
Units: add au_action (hbar) and derived atomic units

mattwelborn:
## Description
This PR adds the atomic unit of action and the derived atomic units on https://en.wikipedia.org/wiki/Hartree_atomic_units. Units are prefixed by `au_`.

## Changelog description
The atomic unit of action and derived atomic units were added.

## Status

- [x] Code base linted
- [x] Ready to go


qcelemental/tests/test_units.py
```diff
@@ -54,6 +54,25 @@ def test_unit_conversion_nist(from_unit, to_unit, expected):
         ("Angstrom", "Angstrom", "1.0"),
         ("coulomb", "statC", "2997924580"),
         ("debye", "e * Bohr", "0.393430273"),
+        # Derived atomic units (https://en.wikipedia.org/wiki/Hartree_atomic_units)
```
dgasmith: +1 for the tests here, especially for the complex units.

qcelemental/physical_constants/ureg.py
```diff
@@ -28,33 +28,53 @@ def build_units_registry(context):
     ureg.define(
         "hartree_inverse_meter = {} / hartree / m".format(phys_const["hartree-inverse meter relationship"]["value"])
     )
+    ureg.define("plancks_constant = {} * joule * s".format(phys_const["planck constant"]["value"]))
```
dgasmith: Did we want `h/hbar` here as well?

qcelemental/physical_constants/ureg.py
```diff
@@ -28,33 +28,53 @@ def build_units_registry(context):
...
+        "au_electric_potential": "atomic unit of electric potential",
+        "au_electric_quadrupole_moment": "atomic unit of electric quadrupole mom.",
+        "au_force": "atomic unit of force",
+        "au_magnetic_dipole_moment": "atomic unit of mag. dipole mom.",
+        "au_magnetic_flux_density": "atomic unit of mag. flux density",
+        "au_magnetizability": "atomic unit of magnetizability",
+        "au_momentum": "atomic unit of mom.um",
+        "au_permittivity": "atomic unit of permittivity",
+        "au_time": "atomic unit of time",
+        "au_velocity": "atomic unit of velocity",
```
dgasmith: What about pressure? I think this is the only one missing.

qcelemental/physical_constants/ureg.py
```diff
@@ -28,33 +28,53 @@ def build_units_registry(context):
...
     )
 
-    # Coulomb
+    # Charge
+    ureg.define("elementary_charge = {} * coulomb = e = au_charge".format(phys_const["elementary charge"]["value"]))
     ureg.define("statcoulomb = coulomb / 2997924580 = statC")
     ureg.define("debye = 1e-18 * statcoulomb * cm = D")
 
+    # Distance
+    ureg.define("bohr = {} * meter = bohr_radius = Bohr = au_length".format(phys_const["bohr radius"]["value"]))
```
dgasmith: What does the following look like? `str(ureg.Unit("5.0 * au_length"))`

qcelemental/physical_constants/ureg.py
mattwelborn: I added `h` and `hbar` to constants in #200 and asked why we have a separate unit registry from the constants. Can we continue this discussion there?

qcelemental/physical_constants/ureg.py
mattwelborn: It's not in the CODATA, at least not under the name "atomic unit of pressure". I will go ahead and add it by derivation.

qcelemental/physical_constants/ureg.py
mattwelborn:
```
In [5]: str(qcelemental.constants.ureg.Unit("5.0 * au_length"))                                                                                                                    
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-5-e84761374df7> in <module>
----> 1 str(qcelemental.constants.ureg.Unit("5.0 * au_length"))

~/anaconda3/envs/qcelemental/lib/python3.7/site-packages/pint/unit.py in __new__(cls, units)
     44             inst._units = units
     45         elif isinstance(units, string_types):
---> 46             inst._units = inst._REGISTRY.parse_units(units)._units
     47         elif isinstance(units, _Unit):
     48             inst._units = units._units

~/anaconda3/envs/qcelemental/lib/python3.7/site-packages/pint/registry.py in parse_units(self, input_string, as_delta)
    790             :class:`ValueError` if the expression is invalid.
    791         """
--> 792         units = self._parse_units(input_string, as_delta)
    793         return self.Unit(units)
    794 

~/anaconda3/envs/qcelemental/lib/python3.7/site-packages/pint/registry.py in _parse_units(self, input_string, as_delta)
    900             as_delta = self.default_as_delta
    901 
--> 902         return super(NonMultiplicativeRegistry, self)._parse_units(input_string, as_delta)
    903 
    904     def _define(self, definition):

~/anaconda3/envs/qcelemental/lib/python3.7/site-packages/pint/registry.py in _parse_units(self, input_string, as_delta)
    810         units = ParserHelper.from_string(input_string)
    811         if units.scale != 1:
--> 812             raise ValueError('Unit expression cannot have a scaling factor.')
    813 
    814         ret = {}

ValueError: Unit expression cannot have a scaling factor.
```

```
In [6]: str(qcelemental.constants.ureg.Unit("au_length"))                                                                                                                          
Out[6]: 'bohr'
```

dgasmith: That wasn't what I was worried about. But is concerning. Let me know if I need to play around with this.

qcelemental/tests/test_units.py
dgasmith: Can we add an `assert str(unit)` to the tests here based off the surprising result below?

qcelemental/physical_constants/ureg.py
```diff
@@ -75,6 +75,8 @@ def build_units_registry(context):
     for k, v in phys_const_map.items():
         ureg.define(f"{k} = {phys_const[v]['value']} * {phys_const[v]['unit']}")
 
+    ureg.define(f"au_pressure = au_energy / au_length**3")  # not in CODATA
```
dgasmith: I suspect it is the fact that this can be derived from other units. CODATA tends not to duplicate.

qcelemental/physical_constants/ureg.py
mattwelborn: I'm not sure I see the issue. Do you want `str(ureg.Unit("au_length")) == "au_length"`? Is it something more?

dgasmith: I see, the first is just my typo. Got it. This is fine.

mattwelborn: I have attempted to fix this with the latest commit (2624d4f), with testing. This does not pass, but looking at why is instructive:

1. some units like "m" are converted to "meter"
2. the amu/kg conversion breaks for some reason

dgasmith:
I don't thing the new commit is needed. It was my misunderstanding originally and the correct behavior seems to be there `au_length -> Bohr`. At least this is my impression of how the conversions should work.

qcelemental/physical_constants/ureg.py
mattwelborn: > I see, the first is just my typo. Got it. This is fine.

In that case, we should probably revert 2624d4f.

mattwelborn:
> I don't thing the new commit is needed. It was my misunderstanding originally and the correct behavior seems to be there `au_length -> Bohr`. At least this is my impression of how the conversions should work.

Okay, I think we are now on the same page. 2627d4f has been reverted.

qcelemental/physical_constants/ureg.py
loriab: Can the unit be singular like the constant? planck, not plancks

EDIT: ah, I see it was like that before, drat.

qcelemental/physical_constants/ureg.py
loriab: codata isn't averse to duplicating but guess they skipped this one.

qcelemental/physical_constants/ureg.py
dgasmith: Maybe add an alias? It should be singular.

dgasmith:
@mattwelborn Merge once "planck" is fixed I believe.

mattwelborn:
Done! But I need an approving review...

qcelemental/physical_constants/ureg.py
dgasmith: Alias should be above `plank_constant = planks_constant = {} * joule...`

Issue #198, opened 2020-01-20
hbar in constants


loriab:
> @dgasmith Also, how would you feel about adding all of the atomic units with generic names (e.g. au_distance, au_charge, au_energy) just to make it easy on people so they don't have to remember Bohr, e, Hartree etc.?

That's strange, I commented on this this afternoon, but it's disappeared. I'll try to remember what I wrote:

* The aliases that are there grandfathered from psi4. I don't object to more (and hbar is a good candidate) but I'd like to gather opinions about whether we should be free or sparing with them.
* Associated is that I aliased the codata 2014 constant names into 2018. how far to we wwant to continue this? 2014 to 2022 or just 2018 to 2022
* Note that codata 2018 changed some units from s to Hz
* Note that the constants haven't been systematically fed into pint. there's some it may not recognize:  GeV, T (period)

mattwelborn:
>     * The aliases that are there grandfathered from psi4. I don't object to more (and hbar is a good candidate) but I'd like to gather opinions about whether we should be free or sparing with them.

I would prefer to be free. I think it's in the spirit of qcel to have as much chemistry (meta)data as possible. 

>     * Associated is that I aliased the codata 2014 constant names into 2018. how far to we wwant to continue this? 2014 to 2022 or just 2018 to 2022

I think we can keep doing this going forward. We only have to do it once every 4 years, and surely they don't change that many names...

>     * Note that the constants haven't been systematically fed into pint. there's some it may not recognize:  GeV, T (period)

I propose we feed the constants into pint (with those constants whose units are not supported being silently ignored)?

dgasmith:
I am for more aliases, the only possible downside here is that we contaminate the space too much making it hard to use. I think this may be hard to do (would need ~5-10x the constants we have now).

As a note on items like GeV, pint understands prefixes:
```
>>> qcel.constants.ureg("GeV")
<Quantity(1, 'gigaelectron_volt')>
```

We might want to explain the Hz thing in the speed of light as it is a bit awkward in code comments.

dgasmith:
Moving further comments here:

@loriab:
> I suspect that most qcel clients are comfortable with direct physical constants and conversion_factor() seems newfangled. So ureg gets most use as user API and isn't so fully built out. I think we want to keep context and ureg separate as they have different sources.

@dgasmith 
> To refine, I think we want to keep the PhysicalConstantsContext class, but slowly deprecated the aliases in favor of direct ureg generated ones?

[...]

PR #200, opened 2020-01-20
Constants: add hbar and fix (?) some units

mattwelborn:
## Description
This PR adds hbar to constants and fixes some of the units. I don't understand:

1. Were the units "incorrect" intentionally?
2. Why do we have constants separate from the unit registry?
3. Why aren't constants in the unit registry?

## Changelog description
hbar was added to constants. The units of c, kb, h, and dipmom_debye2si were corrected.

## Status

- [x] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcelemental/physical_constants/context.py
+++ b/qcelemental/physical_constants/context.py
@@ -148,9 +148,10 @@ class PhysicalConstantsContext:
             ]
 
         aliases.extend([
-            ('h',                    'J',              self.pc['hertz-joule relationship'].data,                             'The Planck constant (Js)'),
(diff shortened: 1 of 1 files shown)


dgasmith:
See conversation in #198, we should continue this there.

loriab:
I haven’t put in tests for codata 2018 yet, so any physconst affecting PR should toggle the line at the end of context.py and check that ureg changes are ok for 2018 as well as 2014.

dgasmith:
To refine, I think we want to keep the `PhysicalConstantsContext` class, but slowly deprecated the aliases in favor of direct ureg generated ones?

This would be my vote anyways so we have a single source of truth.

dgasmith:
Merging, moved conversation to #198.

PR #201, opened 2020-01-22
trigger slim mol in AtomicResult.dict() in test suite

loriab:
## Description

trigger slim mol in AtomicResult.dict() in test suite

should trigger the error:
```
E   pydantic.error_wrappers.ValidationError: 9 validation errors for AtomicResult
E   molecule -> atom_labels_
E     extra fields not permitted (type=value_error.extra)
E   molecule -> atomic_numbers_
E     extra fields not permitted (type=value_error.extra)
E   molecule -> connectivity_
E     extra fields not permitted (type=value_error.extra)
E   molecule -> fragment_charges_
E     extra fields not permitted (type=value_error.extra)
E   molecule -> fragment_multiplicities_
E     extra fields not permitted (type=value_error.extra)
E   molecule -> fragments_
E     extra fields not permitted (type=value_error.extra)
E   molecule -> mass_numbers_
E     extra fields not permitted (type=value_error.extra)
E   molecule -> masses_
E     extra fields not permitted (type=value_error.extra)
E   molecule -> real_
E     extra fields not permitted (type=value_error.extra)
```

uncommenting the `def dict` with `by_alias` yields a new error:
```
E   pydantic.error_wrappers.ValidationError: 2 validation errors for AtomicResult
E   molecule -> masses
E     object of type 'NoneType' has no len() (type=type_error)
E   molecule -> real
E     object of type 'NoneType' has no len() (type=type_error)
```
the skip_defaults fixes that. but this probably isn't the right way (indeed, the skip_defaults line breaks psi4 again)

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcelemental/models/molecule.py
+++ b/qcelemental/models/molecule.py
@@ -501,7 +501,7 @@ class Molecule(ProtoModel):
 
     def dict(self, *args, **kwargs):
         kwargs["by_alias"] = True
-        kwargs["skip_defaults"] = True
+        kwargs["exclude_unset"] = True
         return super().dict(*args, **kwargs)
 
     def pretty_print(self):
--- a/qcelemental/models/molecule.py
+++ b/qcelemental/models/molecule.py
@@ -500,7 +500,8 @@ class Molecule(ProtoModel):
         return self.get_hash() == other.get_hash()
 
     def dict(self, *args, **kwargs):
-        kwargs.setdefault("by_alias", True)
+        kwargs["by_alias"] = True
+        kwargs["skip_defaults"] = True
         return super().dict(*args, **kwargs)
 
     def pretty_print(self):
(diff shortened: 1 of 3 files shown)


dgasmith:
I think this is boolean propagation, we will need to add the `force_xxx` to the Molecule `dict()` interface. `setdefault` isn't cutting it.

loriab:
I'm not following the reference to `force_xxx` args.

dgasmith:
```python
def dict(self, *args, **kwargs):

    kwargs["by_alias"] = True
    kwargs["exclude_unset" ] = True
    return super().dict(*args, **kwargs)
```

So we not longer let the user override. Which is a bit unfortunate.

loriab:
Thanks. got it after I hit the button. This one should pass ...

loriab:
Thanks for the solution. I'd like to get this in soon because I think @dsirianni will hit it. Any test checks beyond not throwing errors?

loriab:
Sure, after group mtg.

PR #203, opened 2020-01-27
add project meta-info CONTRIBUTING.md

loriab:
## Description

Note that I want to keep that very long slack invite link in only one spot, which is README, in case it needs changing again.

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -2,16 +2,16 @@
 
 QCElemental is a project collecting fundamental tools for computational 
 molecular sciences (CMS) into a lightweight Python
-interface. It is matained by the Molecular Sciences Software
+interface. It is maintained by the Molecular Sciences Software
 Institute (MolSSI) and contributed to by a number of community CMS code 
 developers and users.
 
 The project welcomes new contributions and comments!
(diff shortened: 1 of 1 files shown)


dgasmith:
Ran a pass over this.

loriab:
Another pass. Tending toward the generic again, but perhaps that's the right balance.

.github/CONTRIBUTING.md
```diff
@@ -1,3 +1,18 @@
+# QCElemental
+
+QCElemental is a project collecting fundamental tools for computational 
+molecular sciences (CMS) into a lightweight Python
+interface. It is maintained by the Molecular Sciences Software
+Institute (MolSSI) and contributed to by a number of community CMS code 
+developers and users.
+
+The project welcomes new contributions and comments!
+Pull requests are tested and linted to ensure code integrity and quality.
```
mattwelborn: Consider "automatically tested and linted"

loriab: The automatically is true but not sufficient. Since this is the contributing docs, may want to emphasize that you the contributor are going to have to write more tests and prove this is good new code that will benefit downstream, not just pass the existing green checks.

PR #208, opened 2020-02-05
Molecule Hash Canary Tests

dgasmith:
## Description
Adds all Fractal molecule hash canaries to QCElemental and reverts the order of operations change for hashing molecule in the `from_data` procedure. Hashing molecules is fairly sensitive to the order of operations and this was disrupted slightly when the `from_data` procedure stopped passing the `validate` flag through to the `__init__`. This was primarily done to prevent a duplicate to/from schema, but had the worrying effect that the geometry was not correctly rounded. A new flag has been introduced to fix this particular issue.

## Changelog description
Fixes a Molecule hashing issue due to order of operations changes in `Molecule.from_data`.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcelemental/models/molecule.py
+++ b/qcelemental/models/molecule.py
@@ -275,6 +275,8 @@ class Molecule(ProtoModel):
         if validate is None:
             validate = not kwargs.get("validated", False)
 
+        geometry_prep = kwargs.pop("_geometry_prep", False)
+
         if validate:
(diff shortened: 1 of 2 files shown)


dgasmith:
It just shows that we do not have enough tests. Thousands of QCArchive tests, only caught by one!

PR #217, opened 2020-03-03
fix schema for Array classes

loriab:
## Description
This will get us through Molecule, though perhaps it should be still more generic.

## Status

- [x] Code base linted
- [x] Ready to go

PR #218, opened 2020-03-03
GitHub Action btwn QCEl and QCSk

loriab:
## Description
Continues from #204. Generate schema from pydantic qcel for models owned by qcsk, compare them, and issue PR to latter in case change to qcel would affect qcsk. Part of https://github.com/MolSSI/QCSchema/issues/68 effort to consolidate single source of truth here and community discussion there.

* because secrets not passed to PRs from forks, these'll have to be internal PRs. I don't understand how codecov is working from fork PRs.
* I think this is pretty much ready. the whole sequence is:
  * #217 needs to happen here
  * QCSk repo needs to switch to JSON files for each model instead of single python module
  * then the target in the action can switch from loriab/QCSchema to upstream
  * I'm thinking of leaving this PR as-is at the push branch stage, then adding https://github.com/marketplace/actions/github-pull-request-action to QCSk to listen for qcel-* branches

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/.github/workflows/QCSchema.yml
+++ b/.github/workflows/QCSchema.yml
@@ -87,7 +87,7 @@ jobs:
       with:
         directory: ./qcsk
         repository: loriab/QCSchema
-        branch: ${branch}
+        branch: ${{ env.branch }}
         #branch: "qcel-204"
         github_token: ${{ secrets.QCSK_TOKEN }}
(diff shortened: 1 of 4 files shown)


loriab:

PR #221, opened 2020-03-27
Update CI.yml

dgasmith:
## Description
Codecov should now work without specifying keys as the GHA API is now stable and updated!

## Status

- [x] Code base linted
- [x] Ready to go


Lnaden:
Huzzah!

PR #223, opened 2020-04-21
madness molecule w/ test

ahurta92:
## Description

This PR is the first step in adding the ability to run madness from QCA.  It parses molecules into madness input.

## Changelog description

It adds the ability to parse a molecule into something that madness can understand
## Status

- [ ] Code base linted
- [ ] Ready to go


qcelemental/models/align.py
```diff
@@ -22,17 +22,15 @@ class AlignmentMill(ProtoModel):
     """
 
     shift: Optional[Array[float]] = Field(  # type: ignore
-        None, description="Translation array (3,) for coordinates.",
+        None, description="Translation array (3,) for coordinates."
```
loriab, line 25: could you have some odd local black setting? I confirmed that running `make format` doesn't remove the trailing commas for me.

qcelemental/molparse/to_string.py
```diff
@@ -204,7 +205,7 @@ def to_dict(self) -> Dict:
         atom_format = "{elem}"
         ghost_format = "GH"
         # TODO handle which units valid
-        umap = {"bohr": "bohr", "angstrom": "angstroms", "nm": "nanometers", "pm": "picometers"}
+        umap = {"atomic": "atomic", "angstrom": "angstroms", "nm": "nanometers", "pm": "picometers"}
```
loriab, line 208: why change the nwchem block?

qcelemental/molparse/to_string.py
```diff
@@ -226,6 +227,29 @@ def to_dict(self) -> Dict:
             data.keywords["scf__nopen"] = molrec["molecular_multiplicity"] - 1
             data.keywords["dft__mult"] = molrec["molecular_multiplicity"]
 
+    elif dtype == "madness":
+
+        atom_format = "{elem}"
+        ghost_format = "GH"
+        # TODO handle which units valid
+        umap = {"bohr": "atomic", "angstrom": "angstrom"}
```
loriab, line 235: what does "atomic" signify? It sounds more like "expect Cartesian geometry" than "expect Bohr geometry". If the former, fine to hard-code it and not entangle with units. Is there a way to explicitly indicate Bohr even if that's the default?

qcelemental/molparse/to_string.py
ahurta92: sorry this is a mistake

qcelemental/molparse/to_string.py
ahurta92: atomic signifies atomic units.  Madness accepts atomic, a.u. , or au.
I put `{"bohr":"atomic"}` because every other qc code uses Bohr instead of atomic and because when I used "atomic":"a

ahurta92: "atomic":"atomic"  `second_line = f"""units {umap.get(units.lower())}"""` wasn't working

qcelemental/models/align.py
ahurta92: I'm not sure about this.  I forked the most recent version of qcel yesterday and ran make format.

qcelemental/molparse/to_string.py
loriab: ok, makes sense. in umap, the lhs are pint/qcschema units and the rhs are strings the programs use to indicate that unit. so atomic:atomic wouldn't work. if madness accepts `au`, I'd do `"bohr": "au"` as being a little clearer.

qcelemental/models/align.py
loriab: Going to have to return to below to pass
```
    mirror: bool = Field(
        False, description="Do mirror invert coordinates?",
    )
```

loriab: are you sure this doesn't work? my PR testing the linter on the existing with-commas code passed

ahurta92: Yeah, I don't think so... not sure what could be wrong.  Every time I run make format it removes the commas from align.py.  I don't ever remember changing any black settings

PR #233, opened 2020-08-18
provenance default for schema

loriab:
## Description
- [x] provenance switched to dynamic default so qcel doesn't show in schema. still passes mypy
- [x] other small mypy fixes
- [x] necessitates pydantic jump from 1.0 to 1.5

## Status

- [x] Code base linted
- [x] Ready to go


loriab:

PR #237, opened 2020-09-04
QCSchema coordination, testing, and export

loriab:
## Description
This reconciles some details of models that had gotten out of sync with qcsk. It makes the schema generatable from the models. It dumps some examples of schema from the existing test cases. It patches up some mypy errors. Finally, it starts setting up some coordination with qcsk so that PRs to qcel (on upstream) that propose to change the schema get notified over in qcsk, addressing https://github.com/MolSSI/QCSchema/issues/68 . This coordinates with https://github.com/MolSSI/QCSchema/pull/77

## Changelog description

- [x] export models to JSON Schema with `make qcschema`
- [x] generate example json from tests and test against exported schema with `pytest --validate qcelemental/`
- [x] mypy fixes incl. dynamic provenance. bump pydantic to v1.5
- [x] testing configuration and GHA coordination with QCSchema repo
- [x] material changes to models
- [x] description string and formatting changes to models
- [x] a0 as Bohr abbreviation. test basis sets as str or float
- [ ] some GHA details aren't polished yet.
  * to have permissions to push a diff branch to qcsk, model-changing PRs here need to be on upstream, not forks
  * need to be able to detect no-change PRs so they're still allowed from forks
  * I turned off CI checks to non-master to avoid duplicate checks, and now the Required ones aren't satisfied

## Status

- [x] Code base linted
- [ ] Ready to go


loriab:
@dgasmith, a few changes to qcel that you should probably glance over to see if they ring any alarm bells. particularly the first bullet.

* any subtle reason `exclude_none` and `exclude_defaults` weren't options before? https://github.com/MolSSI/QCElemental/pull/237/files#diff-7a9af1eb4f18dc820783bf210d2cfc95
* prov version and routine default going to string https://github.com/MolSSI/QCElemental/pull/237/files#diff-0baf84b1436ced317794f35442380396R22-R26
* `model.basis` can now be full basis https://github.com/MolSSI/QCElemental/pull/237/files#diff-0baf84b1436ced317794f35442380396R44
* prov and error are moving to default_factory https://github.com/MolSSI/QCElemental/pull/237/files#diff-c1df0b88317a0c0331bea59a1793059cR274, https://github.com/MolSSI/QCElemental/pull/237/files#diff-f178293574dc070baaac6713e52806beR495
* time to drop old mp2 name unless that'd be trouble for qcf https://github.com/MolSSI/QCElemental/pull/237/files#diff-f178293574dc070baaac6713e52806beR108
* I added `a0` to pint https://github.com/MolSSI/QCElemental/pull/237/files#diff-a0d915622c0e7d268558ba8b1f5a3216R52 b/c `a_0^2` seems harder to separate than `a0^2`. I'd like `Eh` in addition to `E_h`, too; both are in psi. thoughts?
* it's a little sad that a straight `.json()` from qcel won't validate so need to exclude the `None`s. any troubling implications you see? https://github.com/MolSSI/QCElemental/pull/237/files#diff-37eecce1557470bdbfa6da71286b3090R66

qcelemental/models/__init__.py
```diff
@@ -12,5 +12,18 @@
...
+
+
+def qcschema_models():
+    return [
+        AtomicInput,
+        AtomicResult,
+        AtomicResultProperties,
+        BasisSet,
+        Molecule,
+        Provenance,
```
dgasmith, line 28: `Optimization`?

loriab: Optimization was never added over at qcsk, so I was aiming for a 1:1 changeset. Agree that it'd be good to get OptimizationInput/Result formalized over there soon.

loriab:
> 1. I don't think exclude_defaults was an option at the time.

✔️ 

> 2. Thats fine, but I did like in the past where it forced you to add these.

I don't think it was, though. It's been `None`, not `...`. https://github.com/MolSSI/QCElemental/blob/master/qcelemental/models/common_models.py#L23-L25 The latter I would have left alone as being consistent with qcsk's all three of creator/version/routine being required. Glad to switch to `...` instead of `""`.

> 3. Sounds good, but will require a lot of checks inside QCEngine.

yes, indeed

> 4. Very much needed.

✔️ 

> 5. The mp2 issues isn't linking at the moment, I cannot see what you are referring to.

Maybe this works better https://github.com/MolSSI/QCElemental/blob/master/qcelemental/models/results.py#L77-L79, `mp2_total_correlation_energy`

> 6. The aliases are fine, though I'm not sure E_h will work as expected. I don't remember if there are odd rules around this.

Ok. I haven't explored `Datum` with these aliases either, so was trying for nominal consistency

> 7. The .json link doesn't appear to be working.

`fp.write(instance.json(exclude_unset=True, exclude_none=True))` aka https://github.com/MolSSI/QCElemental/pull/237/files#diff-37eecce1557470bdbfa6da71286b3090R66

dgasmith (approved):
Thanks for the links, it all looks great!

loriab:
It turns out a0/a_0/Eh/E_h are already in pint c. v0.10 https://github.com/hgrecco/pint/pull/811

PR #244, opened 2020-10-11
Models: Minor basis fix if center data is not valid

dgasmith:
Validation would fail with an incorrect error message if `center_data` did not validate in Basis. This PR fixes this issue to always ensure validation errors are thrown at the correct location.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcelemental/models/basis.py
+++ b/qcelemental/models/basis.py
@@ -145,7 +145,12 @@ class BasisSet(ProtoModel):
     @validator("atom_map")
     def _check_atom_map(cls, v, values):
         sv = set(v)
-        missing = sv - values["center_data"].keys()
+
+        # Center_data validation error, skipping
+        try:
+            missing = sv - values["center_data"].keys()
+        except KeyError:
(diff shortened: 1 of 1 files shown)


qcelemental/models/basis.py
```diff
@@ -145,7 +145,12 @@ class BasisSet(ProtoModel):
     @validator("atom_map")
     def _check_atom_map(cls, v, values):
         sv = set(v)
-        missing = sv - values["center_data"].keys()
+
+        # Center_data validation error, skipping
+        try:
+            missing = sv - values["center_data"].keys()
```
loriab, line 151: if there was a pydantic validator on `center_data`, would that do the same trick (since first)?

dgasmith: I do not believe so since pydantic first collects all validation errors then generates the error message. If a validator throws, it interrupts the collection process and throws an incorrect error.

loriab (approved):

PR #246, opened 2021-01-07
Replaces deprecated numpy dtypes

Andrew-AbiMansour:
## Description

Replaces deprecated data types np.float/np.int/np.bool with native float/int/bool.
See [Notes](https://numpy.org/devdocs/release/1.20.0-notes.html) in v1.20.0 on Deprecations.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcelemental/molparse/from_arrays.py
+++ b/qcelemental/molparse/from_arrays.py
@@ -596,7 +596,7 @@ def validate_and_fill_efp(fragment_files=None, hint_types=None, geom_hints=None)
 def validate_and_fill_geometry(geom=None, tooclose=0.1, copy=True):
     """Check `geom` for overlapping atoms. Return flattened"""
 
-    npgeom = np.array(geom, copy=copy, dtype=np.float).reshape((-1, 3))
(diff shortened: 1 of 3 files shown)


loriab:
Thanks very much. I think your changes are fine, but I wish I could read more about: "In many cases, it may have been intended to use the types from the third column. Be aware that use of these types may result in subtle but desirable behavior changes." I'll see if I can find more info.

loriab (approved):
I've tested these changes, and they're fine. And I can't find any more info on the subtle benefits, so let's go with it. I'll know what to suspect if problem turn up. Thanks for the PR!

mattwthompson:
I think a couple were missed - (see https://github.com/mattwthompson/QCElemental/commit/7c5bf8045c8c8aa7ed4e617618de78b1bed561e3) but installing from source removes the deprecations warnings that are printed from the release build, at least for the part of the API we use.

loriab:
whoops, I found a couple more than the initial PR but missed those. go ahead and PR your branch, so as to polish off. thanks.

PR #254, opened 2021-03-04
release notes for v0.19.0

loriab:
## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcelemental/util/autodocs.py
+++ b/qcelemental/util/autodocs.py
@@ -43,8 +43,13 @@ def parse_type_str(prop) -> str:
         fields.SHAPE_SET: "Set",
         fields.SHAPE_LIST: "List",
         fields.SHAPE_SINGLETON: "Union",
-        fields.SHAPE_DICT: "Dict",
     }  # yapf: disable
+    try:
+        typing_map[fields.SHAPE_DICT] = "Dict"
+    except AttributeError:
+        # older pydantic
+        pass
+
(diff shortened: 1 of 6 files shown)


loriab:
@dgasmith, the fields list grew quite a bit https://github.com/samuelcolvin/pydantic/blob/master/pydantic/fields.py#L254-L277, so could you check over the change here, please? `typing_map[12]` was the only error throwing in the test cases.

qcelemental/util/autodocs.py
```diff
@@ -44,6 +44,12 @@ def parse_type_str(prop) -> str:
         fields.SHAPE_LIST: "List",
         fields.SHAPE_SINGLETON: "Union",
     }  # yapf: disable
+    try:
```
dgasmith, line 47: Ah, I was waiting for the 1.8 regressions. I pin pydantic minor these days until 1.x.1 is out, its getting to be a pretty complex system and regardless of their expansive test suite, items filter through.

dgasmith (approved):
LGTM! Nice.

qcelemental/util/autodocs.py
loriab: the big v1.8 one was intended (see breaking: https://github.com/samuelcolvin/pydantic/releases/tag/v1.8). this little typing one was a surprise. Is it supposed to be `Dict` or `Dict[{}]`?

PR #257, opened 2021-03-12
Allow to write Fortran header from codata

### LINKED ISSUE #256: Allow generation of Fortran include or module file from NIST codata

The NIST codata available in qcel can currently be exported to a C header, it would be interesting to have an export to a Fortran include or module file as well.

Have an additional function in https://github.com/MolSSI/QCElemental/blob/master/qcelemental/physical_constants/context.py to export a Fortran include or module file.

**Describe alternatives you've considered**

The C header could be included in Fortran as well since it only contains macro definitions.

**Additional context**

Discussion on parsing NIST codata on the Fortran discourse: https://fortran-lang.discourse.group/t/tips-to-make-this-toy-program-faster/803/38

awvwgk:
## Description

This patch allows to write the constants from the codata as Fortran include file.
Also, allows to adjust the prefix used for writing the C header.

Closes #256 

## Changelog description

Allow writing codata as Fortran include file

## Status

- [x] Code base linted
- [x] Ready to go


loriab (approved):
thank you, lgtm!

loriab:
Let me know when this is ready for merge, otherwise I'll do it tomorrow.

awvwgk:
I put it up for feedback from the Fortran community here: https://fortran-lang.discourse.group/t/physical-constants/822

I think it is ready to go as is, but let's wait until tomorrow for the merge :+1:.

loriab:
Would `1.D-2` be more broadly read than `1.E-2`? Pardon my Fortran, it's 77 vintage.

awvwgk:
The preferred strategy in Fortran is to specify floating point literals as `1.e-2_wp` where `wp` is a kind parameter with a value like `selected_real_kind(15)`. The routine can write as default real kind `1.e-2` which is most likely single precision, but can also take the name of the kind parameter and allow this way to write whatever precision is preferred (single, double, extended double, quadruple).

PR #265, opened 2021-06-02
fixing search path construction

sheepforce:
## Description
When constructing the search path for executables, the string concatenation prepends the pathseparator and append the python executable path to the original search path. Both should be appended. In consequence, the last path in `$PATH` becomes invalid at the moment (and in my case fails to find `define` from turbomole as this is the last directory in my `$PATH`). This PR fixes the construction of the path.

## Changelog description
fixes invalid last entry in search path construnction

## Status

- [ ] Code base linted
- [x] Ready to go

### DIFF
--- a/qcelemental/util/importing.py
+++ b/qcelemental/util/importing.py
@@ -79,7 +79,7 @@ def which(
 
     """
     if env is None:
-        lenv = {"PATH": os.environ.get("PATH", "") + os.pathsep + os.path.dirname(sys.executable)}
+        lenv = {"PATH": os.pathsep + os.environ.get("PATH", "") + os.pathsep + os.path.dirname(sys.executable)}
     else:
         lenv = {"PATH": os.pathsep.join([os.path.abspath(x) for x in env.split(os.pathsep) if x != ""])}
     lenv = {k: v for k, v in lenv.items() if v is not None}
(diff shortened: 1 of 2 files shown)


qcelemental/util/importing.py
```diff
@@ -79,7 +79,7 @@ def which(
 
     """
     if env is None:
-        lenv = {"PATH": os.pathsep + os.environ.get("PATH", "") + os.path.dirname(sys.executable)}
+        lenv = {"PATH": os.environ.get("PATH", "") + os.pathsep + os.path.dirname(sys.executable)}
```
loriab, line 82:
```suggestion
        lenv = {"PATH": os.pathsep + os.environ.get("PATH", "") + os.pathsep + os.path.dirname(sys.executable)}
```

loriab:
fyi, CI error isn't your fault. there was a package realignment upstream, and now py36 pydantic requires dataclasses dep explicitly.

There'll need to be a new qcel release soon to incorporate min pydantic 1.8.2 for security anyway, so this PR's fix can be got in, too.

loriab:
If you add a `dataclasses` line here https://github.com/MolSSI/QCElemental/blob/master/devtools/conda-envs/base.yaml#L12, should heal CI. We'll be dropping py36 soon anyways.

sheepforce:
Done :slightly_smiling_face: I'm absolutely no Python developer so thx for the guidance here. ^^

loriab (approved):
lgtm! thanks for the fix

PR #267, opened 2021-06-29
Add plain msgpack serialization

bennybp:
## Description

Adds ability to serialize to plain msgpack format.

The current `msgpack-ext` encoding serializes numpy arrays as numpy arrays, which is not portable (that is, deserializing would require python and qcelemental). This adds a plain `msgpack` encoding that converts numpy arrays to lists, as is done with the existing `json` encoding.

## Changelog description

Adds plain msgpack serialization

## Status

- [X] Code base linted
- [X] Ready to go


qcelemental/util/serialization.py
```diff
@@ -238,6 +238,78 @@ def json_loads(data: str) -> Any:
...
+    str
+        A msgpack representation of the data in bytes.
+    """
+
+    which_import("msgpack", raise_error=True, raise_msg=_msgpack_which_msg)
+
+    return msgpack.dumps(data, default=msgpack_encode, use_bin_type=True)
+
+
+def msgpack_loads(data: str) -> Any:
```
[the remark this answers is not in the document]
bennybp: Yes. But in the same way that `json_loads` is a duplicate of `jsonext_loads`. I can remove the duplication if you'd like

bennybp:
Added to the init file and fixed the test. I guess I didn't need to fix the test, since `dtype` is inferred by the extension, but it's nice to be explicit I guess.

For the to/from file, I removed the `msgpack-ext` to make it more consistent. So `json-ext` and `msgpack-ext` aren't valid for files.

loriab (commented):
how're these for a sol'n? I'm trying to avoid changing current behavior.

if you date the v0.21 release in changelog, I won't have to make an extra PR

qcelemental/models/molecule.py
```diff
@@ -978,10 +978,8 @@ def to_file(self, filename: str, dtype: Optional[str] = None) -> None:
         flags = "w"
         if dtype in ["xyz", "xyz+", "psi4"]:
             stringified = self.to_string(dtype)
-        elif dtype in ["json"]:
-            stringified = self.serialize("json")
-        elif dtype in ["msgpack", "msgpack-ext"]:
-            stringified = self.serialize("msgpack-ext")
+        elif dtype in ["json", "json-ext", "msgpack", "msgpack-ext"]:
+            stringified = self.serialize(dtype)
             flags = "wb"
```
loriab, line 983:
```suggestion
            flags = "wb" if dtype.startswith("msgpack") else "w"
```

qcelemental/models/molecule.py
```diff
@@ -951,6 +951,10 @@ def from_file(cls, filename: str, dtype: Optional[str] = None, *, orient: bool =
             with open(filename, "rb") as infile_bytes:
                 data = deserialize(infile_bytes.read(), encoding="msgpack")
             dtype = "dict"
+        elif dtype == "msgpack-ext":
+            with open(filename, "rb") as infile_bytes:
+                data = deserialize(infile_bytes.read(), encoding="msgpack-ext")
+            dtype = "dict"
```
loriab, line 957: since for reading json=json-ext and msgpack=msgpack-ext, aren't only 2 if's needed?

bennybp: Yes, although we should make it explicit. Ie, for json, use `json-ext` and for `msgpack`, use `msgpack-ext` (with comment that the `-ext` are backwards compatible)

loriab (approved):

PR #283, opened 2021-11-26
Update `TDKeywords`

jthorton:
## Description
This PR just updates the TDKeywords model to be consistent with the qcfractal model as converting between the two causes issues currently. 

## Changelog description
`TDKeywords` now accepts additional keywords in line with qcfractal.

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcelemental/models/procedures.py
+++ b/qcelemental/models/procedures.py
@@ -190,6 +190,9 @@ class TDKeywords(ProtoModel):
         "start new optimizations, in unit of a.u. I.e. if energy_upper_limit = 0.05, current global "
         "minimum energy is -9.9 , then a new task starting with energy -9.8 will be skipped.",
     )
+    additional_keywords: Dict[str, Any] = Field(
+        {}, description="Additional keywords to add to the torsiondrive's optimization runs."
+    )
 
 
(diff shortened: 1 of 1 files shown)


bennybp:
I've been thinking about this. I'm not sure the additional keywords actually belongs in the TDKeywords. The additional keywords should just be folded into the optimization keywords, right?

I know that was a change we made to QCFractal. Thinking about it, I'm not positive that part was necessary, and need to revisit it for the new version (with input from @dotsdl )

jthorton:
> The additional keywords should just be folded into the optimization keywords, right?

Yeah, that's how it should be done and thats basically what TDKeywords does with the dihedral field. This is just turned into a constraint and passed to the optimisation keywords. I think in QCFractal the additional keywords should be separate from the optimisation for the same reason the dihedral field is. As if this is unique for every entry in a collection of size N this would require N different optimisation specs which only differ by the additional keywords. So it makes more sense to have one consistent optimisation specification and have the additional keywords separate. 

In the case of single jobs via QCengine then its probably fine to keep them separate however I was thinking it might be best to unify all of the models, then users can create a single specification that can be run locally via qcengine or via qcfractal.

PR #284, opened 2021-12-02
Add t-amplitude driver to DriverEnum

cgbriggs99:
## Description

The goal of this pull request is to add a driver to obtain the t-amplitude tensor from a wavefunction. Most of the code for this is found in [https://github.com/psi4/psi4/pull/2328](https://github.com/psi4/psi4/pull/2328).

## Changelog description

Added "tamps" driver to obtain the T-amplitudes from MP2 and CCSD.

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcelemental/models/common_models.py
+++ b/qcelemental/models/common_models.py
@@ -57,6 +57,7 @@ class Model(ProtoModel):
 class DriverEnum(str, Enum):
     """Allowed computation driver values."""
 
+    tamps = "tamps"
     energy = "energy"
     gradient = "gradient"
     hessian = "hessian"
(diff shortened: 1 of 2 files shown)


loriab:
I'm sorry to be discouraging on this, but the model we're trying to stick to is that driver governs the returns on analytic single-points and are pretty limited in number -- hence the e, g, h, properties. I think the best way is to squeeze the instruction into the psi4 properties() cmd (or into extras at a pinch). The amplitudes don't have a schema to adhere to and for other qc programs to match, so they may need to be stored in extras.qcvars.

cgbriggs99:
Thanks for the input. I'll start work on a different approach.

PR #285, opened 2022-03-17
Use empty dictionary for empty native_files to maintain consistency w…

coltonbh:
…ith keywords, extras and other dict objects in models

## Description
`QCElemental's` default behavior has been to include empty dictionaries for `dict` objects on `AtomicInput/AtomicResult` objects. This default allows for nice behavior like checking for the presence of keys of interest with the usual code without having to first check that `keywords` or `extras` is not `None`:

```python
if my_atomic_input.keywords.get('mykey'):
    # do something

if my_atomic_input.extras.get('mykey'):
    # do something
```

The `native_files` field deviates from this standard and requires additional checks to access its fields:
```python
if my_atomic_result.native_files:
    if my_atomic_result.native_files.get("mykey")
        # do something
```
I propose the previous standard be maintained as I think it makes working with the objects cleaner and more consistent :)

## Changelog description
Use empty dictionary for empty `native_files` to maintain consistency with `keywords`, `extras` and other dict objects in models

## Status

- [ x] Code base linted
- [ x] Ready to go

@bennybp lemme know if you have any thoughts :)

### DIFF
--- a/qcelemental/tests/test_model_results.py
+++ b/qcelemental/tests/test_model_results.py
@@ -371,7 +371,7 @@ def test_native_protocols(protocol, provided, expected, native_data_fixture, req
     drop_qcsk(wfn, request.node.name)
 
     if len(expected) == 0:
-        assert wfn.native_files is None
+        assert isinstance(wfn.native_files, dict)
     else:
         expected_keys = set(expected)
         assert wfn.native_files.keys() == expected_keys
--- a/qcelemental/tests/test_model_results.py
+++ b/qcelemental/tests/test_model_results.py
@@ -371,7 +371,7 @@ def test_native_protocols(protocol, provided, expected, native_data_fixture, req
     drop_qcsk(wfn, request.node.name)
 
     if len(expected) == 0:
-        assert isinstance(wfn.native_files, dict)
+        assert wfn.native_files == {}
     else:
         expected_keys = set(expected)
         assert wfn.native_files.keys() == expected_keys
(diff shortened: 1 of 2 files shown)


loriab (approved):
lgtm! Ben's fine with empty dict, too.

coltonbh:
Just adding a note here for future reference on the thinking behind this PR. One of the main ideas behind this suggestion (and I suspect QCElemental's previous standard of empty containers for `.extras` and `.keywords`) is that if a container can be `None` then it has three states (using a dict as an example): `None`, `{}`, `{'key': 'value'}` and the first two mean essentially the same thing and behave similarly in python, with `None` just being a more limited `{}`. The beauty of python evaluating empty containers as falsy (i.e., `assert not {}`) is that you can treat empty containers (`lists`, `sets`, `dicts`) as `None-ish` and then only have two states to deal with: `{}` or `{'key': 'value'}`. This allows pythonic access to the attributes (i.e., ` if atres.extras.get("my_key"): ...`) without having to initially perform checks for a `None` value.

It's been recommended to me that using empty containers for optional values is preferred for containers, whereas for non-container types (`int`, `float`, `str`), `None` is preferred since there is no "empty" `float`/`int` value. String values may be empty (i.e., `""`) and there is debate about which is better (`None` or `""`) and tend to vary by use case, often having to do with how a database handles `NULL` or an empty string.

qcelemental/tests/test_model_results.py
```diff
@@ -371,7 +371,7 @@ def test_native_protocols(protocol, provided, expected, native_data_fixture, req
     drop_qcsk(wfn, request.node.name)
 
     if len(expected) == 0:
-        assert wfn.native_files is None
+        assert isinstance(wfn.native_files, dict)
```
loriab, line 374:
```suggestion
        assert wfn.native_files == {}
```

coltonbh:
Woot woot! 🎉

PR #286, opened 2022-03-24
fix circular import in docs

loriab:
## Description
- [x] Sphinx warns about can't import `ProtoModel` when documenting (with autodoc and typing utilities) `qcel.testing.compare_recursive`. Can't be evaded by import guards b/c Sphinx uses runtime. This works around it and will be for Psi4 docs.
- [x] Fixes some `git+git` to `git+https` since GH doesn't authenticate the former anymore.
- [x] Removes some pytest>=4 restrictions now that pytest has reached 7
- [x] linted
- [x] test py310 and try to remove double testing for branches on upstream

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,6 +1,10 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
 
 jobs:
   build:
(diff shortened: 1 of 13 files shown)


WardLT (approved):
LGTM. Thanks, Lori!

PR #289, opened 2022-06-06
Pydanticdocs

loriab:
## Description
Rework docs (motivation intersphinx from Psi4).

## Changelog description
- [x] Use the excellent https://github.com/mansenfranzen/autodoc_pydantic project to integrate models into sphinx. It looks nice, it is highly configurable, and we can drop the home-grown `AutodocBaseSettings` work that predates this availability (speak up, anyone, if you know reasons to keep it. Here's discussion on docs btwn pydantic and qca folks: https://github.com/samuelcolvin/pydantic/issues/638 .
- [x] More classes/fns documented and API section used as superset, not complement to non-auto docs.
- [x] Fixed a couple things `make linkcheck` didn't like.
- [x] Added links to models.
- [x] Retired the specialized qca sphinx theme in favor of rtd. There was some awkward formatting by qca.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcelemental/models/results.py
+++ b/qcelemental/models/results.py
@@ -568,15 +568,16 @@ class AtomicInput(ProtoModel):
             f"The QCSchema specification this model conforms to. Explicitly fixed as {qcschema_input_default}."
         ),
     )
-    schema_version: int = Field(1, description="The version number of :attr:`~qcelemental.models.AtomicInput.schema_name` to which this model conforms.")
+    schema_version: int = Field(
+        1,
+        description="The version number of :attr:`~qcelemental.models.AtomicInput.schema_name` to which this model conforms.",
+    )
 
     molecule: Molecule = Field(..., description="The molecule to use in the computation.")
     driver: DriverEnum = Field(..., description=str(DriverEnum.__doc__))
     model: Model = Field(..., description=str(Model.__doc__))
     keywords: Dict[str, Any] = Field({}, description="The program-specific keywords to be used.")
-    protocols: AtomicResultProtocols = Field(
-        AtomicResultProtocols(), description=str(AtomicResultProtocols.__doc__)
-    )
+    protocols: AtomicResultProtocols = Field(AtomicResultProtocols(), description=str(AtomicResultProtocols.__doc__))
 
     extras: Dict[str, Any] = Field(
         {},
@@ -762,6 +763,7 @@ class ResultProperties(AtomicResultProperties):
        Use :py:func:`qcelemental.models.AtomicResultProperties` instead.
 
     """
+
     def __init__(self, *args, **kwargs):
         from warnings import warn
 
@@ -779,6 +781,7 @@ class ResultProtocols(AtomicResultProtocols):
        Use :py:func:`qcelemental.models.AtomicResultProtocols` instead.
 
     """
+
     def __init__(self, *args, **kwargs):
         from warnings import warn
 
@@ -796,6 +799,7 @@ class ResultInput(AtomicInput):
        Use :py:func:`qcelemental.models.AtomicInput` instead.
 
     """
+
     def __init__(self, *args, **kwargs):
         from warnings import warn
 
@@ -810,6 +814,7 @@ class Result(AtomicResult):
        Use :py:func:`qcelemental.models.AtomicResult` instead.
 
     """
+
     def __init__(self, *args, **kwargs):
         from warnings import warn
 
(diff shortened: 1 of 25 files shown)


qcelemental/models/results.py
```diff
@@ -287,7 +288,7 @@ def _validate_derivs(cls, v, values, field):
         try:
             v = np.asarray(v).reshape(shape)
         except (ValueError, AttributeError):
-            raise ValueError(f"Derivative must be castable to shape {shape}!")
+            raise ValueError(f"Derivative must be castable to shape {shape}! len={np.asarray(v).size()}")
```
Lnaden, line 291: Can this cause issues if `v` isn't something which can be reasonably cast through `asarray`?

qcelemental/molutil/align.py
```diff
@@ -554,34 +554,41 @@ def kabsch_quaternion(P, Q):
     return U
 
 
-def compute_scramble(nat, do_resort=True, do_shift=True, do_rotate=True, deflection=1.0, do_mirror=False):
+def compute_scramble(
+    nat: int,
+    do_resort: Union[bool, List, np.ndarray] = True,
```
Lnaden, line 559: Very minor, but tuple would also be valid since these are all technically `array-like`. Would it be better to be a `Union[bool, numpy.typing.ArrayLike]` as a more complete catch-all? Same for the other 2 here as well.

qcelemental/molutil/connectivity.py
```diff
@@ -16,18 +16,18 @@ def guess_connectivity(
...
+    threshold
         Tunes the covalent radii metric safety factor.
-    default_connectivity : Optional[float], optional
+    default_connectivity
         Provides a default connectivity value
 
     Returns
     -------
-    List[Union[Tuple[int, int], Tuple[int, int, float]]]
+    ~typing.List[~typing.Union[~typing.Tuple[int, int], ~typing.Tuple[int, int, float]]]
```
Lnaden, line 30: Do we need `~typing...` here? They were stripped out of other modules so I want to make sure we're not adding extra bulk.

Lnaden (approved):
Overall look good. There is one functional issue in error handling I raised which should be looked at.

Generally, is there a reason all the type information was stripped from the parameter name in the function docstring? E.g.

`path : Union[str, Path]` -> `path`?

qcelemental/molutil/align.py
loriab: ooh, I didn't know numpy supplied `ArrayLike` -- I had only seen `array-like` for human consumption. thanks!

Lnaden: I didnt either! I knew there was something in the `collections` lib, but in looking while reviewing, I found that. Its only available from 1.20+ though, not sure if thats going to be a barrier.

qcelemental/molutil/connectivity.py
loriab: I can answer that with your `path : Union[str, Path] -> path` query.

So with newer sphinx auto/numpy/etc extensions, sphinx can entirely use typing info for parameters from the function signature. having it also in the docstring only overrides the signature typing and changes up the formatting for the worse. so no need for it in docstrings anymore.

If you put the return typing in the signature and don't need a comment for it, the autodocs can use the signature return type. If you do want a docstring comment for the return, the return type does have to be repeated in the docstring to be compatible with numpy style. you can have plain `Union[str, Tuple]` there, but the Union and Tuple won't be linked back to python docs as they normally would. `typing` allows them to be a linked and `~` suppresses the typing.

qcelemental/models/results.py
loriab: Yeah, info may not be worth the risk.

Lnaden: If you want info, you can just post `v` so someone can see what the problematic entry was?

qcelemental/molutil/align.py
loriab: I'm trying to keep qcel fairly undemanding (py36+, numpy from years back).
The below can work, but the results (w/o, w) are harder to read and less linked. So I'll stick with List, ndarray for now. Glad to know about it, though!
```
try:
    from numpy.typing import ArrayLikeMine
except ImportError:
    ArrayLike = Union[List, np.ndarray]
```

Screen Shot 2022-06-07 at 3 29 18 PM
Screen Shot 2022-06-07 at 3 28 14 PM

PR #299, opened 2022-11-11
noncontig passthrough

### LINKED ISSUE #298: Setup qcel.models.Molecule with non-contiguous fragments
Trying the following example molecule from S22, I notice that there is something odd happening for molecules with fragment information:

````python
import qcelemental as qcel

data = {
    "symbols": [
      "N", "N", "H", "H", "H", "H", "H", "H"
    ],
    "geometry": [
      -2.9833455085754399E+0, -8.8082052767279997E-2,  0.0000000000000000E+0,
       2.9833455085754399E+0,  8.8082052767279997E-2,  0.0000000000000000E+0,
      -4.0792036056518599E+0,  2.5775116682053001E-1,  1.5298565626144400E+0,
      -1.6052680015564000E+0,  1.2438048124313399E+0,  0.0000000000000000E+0,
      -4.0792036056518599E+0,  2.5775116682053001E-1, -1.5298565626144400E+0,
       4.0792036056518599E+0, -2.5775116682053001E-1, -1.5298565626144400E+0,
       1.6052680015564000E+0, -1.2438048124313399E+0,  0.0000000000000000E+0,
       4.0792036056518599E+0, -2.5775116682053001E-1,  1.5298565626144400E+0
    ],
    "fragments": [
        [0, 2, 3, 4],
        [1, 5, 6, 7],
    ],
}

for validate in (True, False):
    try:
        mol = qcel.models.Molecule(**data, validate=validate)
    except qcel.exceptions.ValidationError as e:
        print(f"Fails to setup with validate={validate}\n{str(e)}")
````

Output from the run is an unexpected validation error when validating the input

````
Warning: QCElemental is reordering atoms to accommodate non-contiguous fragments
Fails to setup with validate=True
Error: QCElemental would need to reorder atoms to accommodate non-contiguous fragments
````

Why are non-contiguous fragments a validation error in qcel? The issue is probably that `throw_reorder` is always true at:

https://github.com/MolSSI/QCElemental/blob/292350f4cc06ca274a15b5e7101e56f25b0093a9/qcelemental/molparse/from_schema.py#L57

while the default is false at

https://github.com/MolSSI/QCElemental/blob/292350f4cc06ca274a15b5e7101e56f25b0093a9/qcelemental/molparse/from_schema.py#L103

but there seems to be no way to toggle `throw_reorder` other than skipping the validation step.

loriab:
## Description
Tentative allowing noncontiguous fragment _input_ in Molecule _with_ validation. Partially fixes #298.

Background is that Mol can store noncontig fragments. However, the internals that the validation step uses can't store them; it can either balk or reorder the atoms so the fragments are contiguous. This PR makes the latter option accessible through the Mol constructor.

This option should be used tentatively and with results checking at first. Note that identical atom ordering is not assured with `allow_noncontiguous_but_reorder=T` between `validate=T/F`. Those wanting validation _and_ atom order presevation may want to `Molecule(..., allow_noncontiguous_but_reorder=True)`, and if that passes, use the results from `mol = Molecule(..., validate=False)`.

This is not a permanent solution, and a future solution with some sort of `fix_order` (like `fix_com`; remember, "affix", not "correct") will use a different option, since reordering that is unnecessary for the schema spec is not good.

## Status

- [x] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcelemental/models/molecule.py
+++ b/qcelemental/models/molecule.py
@@ -341,14 +341,22 @@ class Molecule(ProtoModel):
             # original_keys = set(kwargs.keys())  # revive when ready to revisit sparsity
 
             nonphysical = kwargs.pop("nonphysical", False)
+            throw_reorder = not (kwargs.pop("allow_noncontiguous_but_reorder", False))  # experimental
             schema = to_schema(
-                from_schema(kwargs, nonphysical=nonphysical), dtype=kwargs["schema_version"], copy=False, np_out=True
+                from_schema(kwargs, nonphysical=nonphysical, throw_reorder=throw_reorder),
+                dtype=kwargs["schema_version"],
+                copy=False,
+                np_out=True,
             )
             schema = _filter_defaults(schema)
 
             kwargs["validated"] = True
             kwargs = {**kwargs, **schema}  # Allow any extra fields
             validate = True
+        else:
+            # these kwargs only relevant for validation
+            kwargs.pop("nonphysical", None)
+            kwargs.pop("allow_noncontiguous_but_reorder", None)
 
         super().__init__(**kwargs)
 
--- a/qcelemental/molparse/from_schema.py
+++ b/qcelemental/molparse/from_schema.py
@@ -7,7 +7,7 @@ from ..util import provenance_stamp
 from .from_arrays import from_arrays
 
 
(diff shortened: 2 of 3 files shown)


awvwgk (approved):

awvwgk:
Anything blocking this from getting merged?

PR #304, opened 2023-02-23
Mol descriptors

dr-marsmm:
## Changelog description

Add MolecularDescriptors class with descriptors as attributes for descriptor attribute in AtomicResultProperties

## Status

- [X] Code base linted
- [X] Ready to go


loriab:
Are these rdkit-specific?

I'm thinking they're a little too specific for the general molecule class. Have you reconsidered since first opening the PR, or would they still be useful, @dr-marsmm?

Issue #303, opened 2023-02-15
Molecule declares extras dictionary as Dict[str, Any] but value can be None


awvwgk:
**Describe the bug**

The `Molecule` class declares the `extras` dictionary as `Dict[str, Any]` but initializes its value with `None`.

**To Reproduce**

````python
>>> import qcelemental as qcel
>>> mol = qcel.models.Molecule(["H", "H"], [[0.0, 0.0, 0.0], [0.0, 0.0, 1.0]])
>>> mol.extras
None
````

**Expected behavior**

An empty dictionary if `extras` are not initialized.

**Additional context**

An easy workaround is
````python
>>> extras = mol.extras if mol.extras is not None else {}
````

PR #305, opened 2023-03-24
Make sure extras is initialized as dict for Molecule

awvwgk:
## Description

Closes https://github.com/MolSSI/QCElemental/issues/303

## Changelog description

Ensure extras attribute in Molecule is initialized as dict

## Status

- [x] Code base linted
- [x] Ready to go


loriab (commented):
Thanks. I'd ordinarily have put the empty dict here https://github.com/MolSSI/QCElemental/blob/master/qcelemental/models/molecule.py#L293 instead of the init. According to https://stackoverflow.com/a/73621352 , it's handled correctly. 

But as-is should be fine.

loriab (approved):

PR #307, opened 2023-03-24
Added ability for .json() to accept kwargs. model.json() can now crea…

coltonbh:
…te a pretty string with .json(indent=4)

## Description
The current implementation of `serialize` does not accept generic `kwargs`, so we loose the basic ability to modify serialized output, like pretty printing a json string with `model.json(indent=4)`. I'm implemented serialize to be slightly more generic so we maintain functionality described in the [pydantic docs](https://docs.pydantic.dev/usage/exporting_models/#modeljson)

## Changelog description
Added ability for .json() to accept kwargs. model.json() can now create a pretty string with .json(indent=4)

## Status

- [x ] Code base linted
- [ x] Ready to go

### DIFF
--- a/qcelemental/models/basemodels.py
+++ b/qcelemental/models/basemodels.py
@@ -127,6 +127,7 @@ class ProtoModel(BaseModel):
         exclude_unset: Optional[bool] = None,
         exclude_defaults: Optional[bool] = None,
         exclude_none: Optional[bool] = None,
+        **kwargs,
     ) -> Union[bytes, str]:
         r"""Generates a serialized representation of the model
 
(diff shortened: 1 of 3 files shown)


coltonbh:
Tagging @loriab and @bennybp for a review. Thanks guys! The CI format and build are failing for reasons unrelated to my PR.

qcelemental/models/basemodels.py
```diff
@@ -127,6 +127,7 @@ def serialize(
         exclude_unset: Optional[bool] = None,
         exclude_defaults: Optional[bool] = None,
         exclude_none: Optional[bool] = None,
+        **kwargs,
```
loriab, line 130: My only hesitation is because I don't know if serialize was written w/o kwargs as a defensive measure. Certainly the indentation use to which you've put kwargs is reasonable and harmless, though. @bennybp, any complications to the database if formats can change through this?

bennybp: I don't believe this will cause any problems on my end

PR #310, opened 2023-03-28
Modernize DevOps Tooling

coltonbh:
## Description
Modernize DevOps tooling. No changes to code functionality in qcel, only devops tooling.

## Changelog description
- Added `/scripts` directory to root of project that contains scripts for testing, formatting code, and building docs.
- Updated build system from `setuptools` to modern `pyproject.toml` specification using `poetry` for the build backend.
- Removed complicated versioning code in favor of single source of truth in `pyproject.toml`. Using standard library `importlib` for looking up package version in `__init__.py` file. 
- Added `build_docs.sh` script to `/scrips` and removed `Makefile` from `/docs`. Flattened `/docs` file structure.
- Removed `travis-ci` code from `devtools`
- Removed LGTM code (they no longer exist as a project).

## TODO
- Figure out what the `pytest --validate` thing was about. @loriab said this was OK to drop.

## Did Not Do

- Test currently write scratch data to the repo in to `qcelemental/tests/qcshema*/*.json`. The tests rely upon `/qcschema/{NAME}/` already being created, so these currently junk up the repo with a `dummy` file in them so git will keep the empty folders in the repo. Modify tests to remove these unnecessary directories to clean up the repo. Test should _never_ write data to the repo. Junk data should be written to `/tmp` or some other temporary in memory buffer. If we want this output data for something (like QCSChema examples) we should have a `scripts/build_examples.sh` script, not couple the task to our test. Separation of concerns :)
- Tests that call PubChem should not. Tests should never rely upon external services (like web APIs). We should monkeypatch http calls to return the data as if PubChem were returning it.

## Notes
- The max version limitations in dependencies are due to packages no longer supporting python3.7, so we have to specifiy max versions. We can remove these limitations once python3.7 reaches end of life June 27, 2023.

## Status

- [x ] Code base linted
- [ x] Ready to go


loriab:
I can see the user-concern/developer-concern distinction -- certainly I wouldn't run MS Word tests, and I've maybe run Python tests once. But those aren't so close to research and probably get much broader testing at release time than qca.

Sure, qcel is tested before pkg building :-) For my part, it's likely even been tested a couple layers downstream through qcng, psi4, qcdb (nwc, gms, cfour). But that's still not great environmental coverage. For example, devs test may on Linux, but maybe we don't generalize the paths right, so something breaks for the user on Windows. Or devs test with MKL, but users run with OpenBLAS, or worse half-MKL, half-OpenBLAS (thanks, pypi). Or devs test with latest and min-version dependencies, and an unintentional breaking change is released for a dep the next day. Or devs test with released versions and users have development versions of several packages. Users are great at installing software combos that faintly horrify developers.

Most of these troubles do involve complications from the compiled code and library compatibility tangles you mentioned, not qcel itself. And some of these arguments apply less strongly to qcel than to qcengine, which is literally tying together these unwieldy projects that never expected to bump shoulders. But tests are fairly light and a great way to diagnose package vs. environment problems. Since qcarchive packages are closer to research than broader tools like fastapi, numpy is probably a better model, and it looks like it packages its tests.

coltonbh:
I hear you :)

A few additional thoughts:

1. `numpy` packages its tests because there is a `build` step associated with installing it (at least if you use a `srcdist` instead of a `wheel`). You have to perform a _compilation_ step with `numpy`, and that can have issues (MKL/BLAS stuff is common, as you point out). Packages that require builds using local compilers are a separate class of code. We are pure python.
2. I'd like to change your perspective that `qcel` is research code! It's a  very basic data structures package in pure python. Sure, it's _used_ in research--but this isn't research code. It's a very standard python library :)
3. Re: your points above, if something is broken for end users as you mention--like a Windows user--having tests on their machine won't help anything. The package already broke for them and they have the corresponding stack trace for opening an issue. Test will not run either, for the same reason. What's the value that running broken tests adds? Breaking changes in newly released dependencies is our fault--it means we didn't specify dependencies correctly in the `pyproject.tom`. The '^1.3.4` notation you see in there now means only major version `1.x.x` allowed, so it avoid the broken scenarios where a new `2.0.0` version breaks code. Again, running tests can't help these end users--test are broken too in this case. As for mixed dependencies, that's on package managers like `poetry`, `pipenv`, etc. to root out--presumably if the versions are such that the package doesn't work, the test won't work either because they use the package! :)
4. Lastly, imagine how end users would run test. The `pip install qcelemental` and they get test code. But can they run it? No, because the testing dependencies are not installed. So now we're distributing non-running code. How do they discover dependencies for tests? They'd look at `pyproject.toml` in the git repo. How would they install them? `pip install pk1, pk2, etc.`. Or more simply, they'd do this:
    ```sh
    git clone qcelemental-git-url.ssh
    poetry install
    pytest
    ````
    End users _can_ always run tests. And the easiest, fastest way to run them is to clone the repo, install all deps for the package and testing, and then run the test. Much easier than `pip install qcelemental`, then manually adding all the deps for tests, then running the tests from your pip installed `qcel`, which most users probably won't even know where it is (would need to run `pytest /path/to/some/python/lib/python3.7/site-packages/qcelemental` and I doubt most users will be able to track that down.

**That said, you are the boss on this! If you want tests in the distributions, I'll add them back.**

I would propose the follow solution as I think it provides the best end-user experience for what you have in mind. Though again, if their package is broken the test can't run--so I don't see the extra value ;)
1. I'd change all testing dependencies as `extras` package dependencies so end users can `pip install qcelemental[test]` and get the test code _and_ required dependencies to run.
2. End users _still_ need to know where they qcelemental package lives. They can find it with the following:
    ```py
    import qcelemental
    print(qcelemental.__file__)
    ````
    Then they can run the tests with `pytest /path/to/qcelemental/install/location`

**What would you like done?**

qcelemental/util/importing.py
```diff
@@ -95,22 +95,3 @@ def which(
...
-    """
-    Package resources is a very slow load
-    """
-    import pkg_resources
-
-    version = pkg_resources.safe_version(*args, **kwargs)
-    return version
-
-
-def parse_version(*args, **kwargs):
```
loriab, line 110: I'm pretty sure these fns are used downstream. Are they deprecated or deprecated from pkg_resources specifically?

raw_data/nist_data/build_periodic_table.py
```diff
@@ -5,10 +5,8 @@
 import datetime
 import json
 import re
-from decimal import Decimal
 
 import requests
-from yapf.yapflib.yapf_api import FormatCode
```
loriab, line 11: This can be commented, but best to have it available (at least in comments) to match formatting if the FormatCode line at end of file is revived. Sorry, GH won't let me add it back as suggestion.

pyproject.toml
```diff
@@ -1,3 +1,88 @@
+[build-system]
+requires = ["poetry-core"]
+build-backend = "poetry.core.masonry.api"
+
+
+[tool.poetry]
+name = "qcelemental"
+version = "0.25.1"
```
loriab, line 8: So now version is decoupled from `git describe` and tags? Possibly that's ok, as we don't end up testing or pinning on non-tagged versions, but it's a little awkward to not be able to tell clean/dirty installations. Also now multiple sources of truth between pyproject.toml and git tags. What was the reasoning behind this vs. versioneer vs. one of https://github.com/python-versioneer/python-versioneer#similar-projects ?

loriab (commented):
Thanks for all the modernizations. This is a partial review with a few questions.

qcelemental/checkup_data/cfour_primary_masses.py
```diff
@@ -1,118 +1,117 @@
 # DATA ATMSS from ~line 150 of cfour/joda/pertable.f
 
-# fmt: off
```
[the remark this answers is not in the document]
coltonbh: Fair. I could add it back. The thinking was to move towards a fully `flake8` compliant repo and it's easier if everything is formatted rather than having to annotate for formatters and linters. Just less "special cases" throughout the code. Can add back `# fmt:off` stuff if you'd like. Let me know.

CONTRIBUTING.md
```diff
@@ -16,24 +16,26 @@ We welcome contributions from external contributors, and this document describes
...
+  poetry install
   ```
 
-- Install `QCElemental` into this virtual environment and then source the new command line tools. The command below installs QCElemental in "editable" mode--meaning all changes to the code will be instantly available to the python interpreter--and with all associated development tools.
+- Activate your new virtual environment. Many editors--like VS Code--will do this for you automatically when you open a directory that has been installed with `poetry`.
 
   ```sh
-  pip install -e '.[docs,tests,align,viz,lint]'
-  source env/bin/activate
+  poetry shell
```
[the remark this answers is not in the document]
coltonbh: `poetry` is just a development tool. Users would always install qcelemental with `pip install qcelemental`. End users would never install or use `poetry`. They can install from source like always with `pip install https://github.com/MolSSI/QCElemental.git@branch` or with `pip install .` if they have the repo locally. `poetry` doesn't define any limitations on what you can do with a `pyproject.toml` repo--this is all related to PIP 517/518 to move the community away from `distutils` and `setuptools`.

You can still install the project into arbitrary directories with `pip install --target=/path/to/directory/ .`

pyproject.toml
coltonbh: Complexity and standard practice. `versioneer` is leaning hard on the single-source-of-truth principle and I like where they're head is at, but it's over-engineered and unnecessary IMO. `Versioneer` been around for 12 years, yet has <1k stars--it's never really taken off. We were seeing bugs on the test suite from _version_ tests. We even had to write version tests! That's a lot of overhead for return a string. Even large python projects get by with a simple string in `__init__.py` (or `pyproject.toml`). I figured we would too :)

qcelemental/util/importing.py
coltonbh: Added back. I see. You mean people use these functions for looking up _other_ pacakge versions, I presume.

CONTRIBUTING.md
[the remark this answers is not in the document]
coltonbh: Also, you would never run `poetry` in a conda environment. `poetry` is your environment manager. It doesn't make sense to run two at a time. If `qcel` goes into a conda environment it's via `conda install` or `pip install`, never `poetry`.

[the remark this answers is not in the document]
coltonbh: Let me know if you have more questions re: poetry. It's just a lightweight environment manager and `pyproject.toml` specification interface for _development_ purposes.

qcelemental/checkup_data/cfour_primary_masses.py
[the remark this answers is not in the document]
coltonbh: Reverted files.

qcelemental/models/__init__.py
```diff
@@ -13,10 +13,16 @@
...
-from .results import Result, ResultInput, ResultProperties  # scheduled for removal
+from .procedures import OptimizationInput, OptimizationResult
+from .results import (  # scheduled for removal
+    AtomicInput,
+    AtomicResult,
+    AtomicResultProperties,
+    Result,
+    ResultInput,
+    ResultProperties,
+)
```
loriab, line 25:
```suggestion
from .results import (
    AtomicInput,
    AtomicResult,
    AtomicResultProperties,
    Result,  # scheduled for removal
    ResultInput,  # scheduled for removal
    ResultProperties,  # scheduled for removal
)
```

qcelemental/util/__init__.py
```diff
@@ -1,10 +1,8 @@
 from .autodocs import auto_gen_docs_on_demand, get_base_docs
 from .gph_uno_bipartite import uno
-from .importing import parse_version, safe_version, which, which_import
+from .importing import which, which_import
```
loriab, line 3: I think we need to keep these fns, as they're used downstream https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/rdkit.py#L8

qcelemental/tests/test_importing.py
```diff
@@ -147,13 +147,3 @@ def test_which_f_raisemsg():
         qcel.util.which("evills", raise_error=True, raise_msg="Install `evills`.")
 
     assert str(e.value).endswith("Command 'evills' not found in envvar PATH. Install `evills`.")
-
-
-def test_safe_version():
-    assert "v" + qcel.util.safe_version(qcel.__version__) == qcel.__version__
-
-
-def test_parse_version():
```
loriab, line 156: these perhaps stay, too, since the functions do.

loriab (commented):
lgtm

qcelemental/util/__init__.py
coltonbh: Fixed!

coltonbh:
@loriab updates made. Test passing! Review pls?

loriab (approved):

coltonbh:
Woot woot! <3

PR #311, opened 2023-05-17
Update docs location

loriab:
## Description
qcel hasn't used ReadTheDocs in some time, and I guess the MolSSI qca website isn't redirecting properly to the working GH Pages docs. I get tired of digging for them, so I changed the site link on GH, and this takes care of the readme. I've also put in a PR to the qcel feedstock (we'll wait for more changes there). Back to the psi4 docs ...

## Status

- [ ] Code base linted
- [ ] Ready to go


Lnaden (approved):
Huzzah! This looked like it was a pain for the dependency shenanigans.

loriab:
Yeah, scipy is an extra dimension to version resolving. Probably we had it in the conda envs but not in setup.py, so Colton didn't notice it for the poetry transition. @Lnaden 

Also, since we're mostly all new to poetry, the "extras" changes come about because from what I gather from https://python-poetry.org/docs/pyproject/#extras, "extras" are for user subsets for optional add'l capabilites (like viz and align) whereas groups are for sets of development dependencies. The "extras" arg where it was before this PR was actually to pull the "viz" extra from package networkx, not assigning networkx to the qcel "viz" extra. So these changes cut down on the default deps installed (back to numpy, pint, pydantic) unless a `poetry install ... --extras ...` gets invoked; then py3dmol, networkx, etc. start getting installed.

Pinging @coltonbh in case I've got this all wrong. :-)

But pretty much I'll merge this, make a v0.26.0 tag (even though poetry probably doesn't use it), do the c-f feedstock, then on to qcng (which may in turn necessitate a v0.26.1 here).

coltonbh:
> Yeah, scipy is an extra dimension to version resolving. Probably we had it in the conda envs but not in setup.py, so Colton didn't notice it for the poetry transition. @Lnaden
> 
> Also, since we're mostly all new to poetry, the "extras" changes come about because from what I gather from https://python-poetry.org/docs/pyproject/#extras, "extras" are for user subsets for optional add'l capabilites (like viz and align) whereas groups are for sets of development dependencies. The "extras" arg where it was before this PR was actually to pull the "viz" extra from package networkx, not assigning networkx to the qcel "viz" extra. So these changes cut down on the default deps installed (back to numpy, pint, pydantic) unless a `poetry install ... --extras ...` gets invoked; then py3dmol, networkx, etc. start getting installed.
> 
> Pinging @coltonbh in case I've got this all wrong. :-)
> 
> But pretty much I'll merge this, make a v0.26.0 tag (even though poetry probably doesn't use it), do the c-f feedstock, then on to qcng (which may in turn necessitate a v0.26.1 here).

Yes, @loriab is correct. Extras are for _end user_ extras--usually augmented features of your package that you don't want to install by default, often because they are heavy-weight or only used for special cases, but want to make available to end users. You can install them with the usual `pip` extras syntax: `pip install qcelemental[viz,whatever]`. On some systems you need to put a `''` around the packages depending on the terminal, e.g., `pip install 'qcelemental[viz,whatever]'`.

You'd never want end users to install your developer dependencies--pytest, isort, etc.--and these are placed in the `dev` group of dependencies. If you have a complex package you can create additional dependency groups for various development environments.

To summarize:
1. All end user dependencies go under [tool.poetry.dependencies]. End user extras are listed as such with the `extras` tag and will be installable by end users from `pypi` (or source) with the `pip install mypkg[extra1,extra2]` syntax.
2. Groups are for _development_ purposes and cannot be accessed by end users from pypi or source using standard `pip install` syntax. They allow for more complex development environment setup. For our use case, having all dev dependencies in a single group is fine.

PR #314, opened 2023-07-28
Use Pydantic v1 API even with pydantic v2

Lnaden:
## Description
This PR changes all imports to test against Pydantic v2 and then uses the v1 api if found. Precursor to future v2 migration.

Also updates the CI to test against both versions. Hopefully I figured out Poetry correctly.

## Changelog description
Compatibility with Pydantic v2 but use the v1 API as a stopgap for the version pin.

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -12,6 +12,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.7", "3.11"]
+        pydantic-version: ["1", "2"]
 
     steps:
       - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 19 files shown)


loriab:
I like your `import pydantic.v1` plan to allow more pydantic versions.

From https://stackoverflow.com/a/70632360 , I think your attempts to manipulate poetry in the GHA through pyproject.toml groups might not be allowed b/c extras not exclusive. Poetry seems ... uncooperative ... for this purpose.

I wonder if a workaround is to return pyproject.toml to `pydantic >=1.8.2` (approx.). Then for the GHA with pydantic=1 in build matrix, do a `sed` on pyproject.toml to turn it to `pydantic >=^1.8.2`. Might work?

Lnaden:
The `import pydantic.v1` was something they added to the v2 of pydantic in case you wanted the API from v1 still, so its not long term stable, but stable enough.

I was worried about the Poetry part because it does say it'll resolve all optional dependencies in the tree too, which really makes me wonder why they don't allow superseding versions of dependencies such as for dev versions.

> Then for the GHA with pydantic=1 in build matrix, do a sed on pyproject.toml to turn it to pydantic >=^1.8.2. Might work?

I'm willing to try something to this effect, I just wont have time to get to it this weekend, so don't let this PR get in the way of other PR's/releases you want to do.

Lnaden:
Okay, this should be ready to go! Failing error is not related to the functional code (just the doc building which is also pydantic problems but not in the way of using the code)

@loriab the `sed` idea worked great, it does a replacement for checking against v1. I left my 'what the hell does this sed string do' comments in place in cased someone else has to edit it that isn't me.

loriab (approved):
Hooray, thanks for finding that `pydantic.v1` workaround so we don't have to be so restrictive. I'll give the build error a try elsewhere.

Lnaden:
Thats actually a great idea. I've added the `sed` here for the `build_documentation` directive, I think that will resolve all the issues and we can re-touch on #311 after.

PR #315, opened 2023-08-01
Don't change global state when hashing molecule

awvwgk:
## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcelemental/models/molecule.py
+++ b/qcelemental/models/molecule.py
@@ -768,7 +768,6 @@ class Molecule(ProtoModel):
         m = hashlib.sha1()
         concat = ""
 
-        np.set_printoptions(precision=16)
         for field in self.hash_fields:
             data = getattr(self, field)
             if field == "geometry":


loriab:
does this need to be a 0.26.1 patch release?

qcelemental/models/molecule.py
```diff
@@ -768,7 +768,6 @@ def get_hash(self):
         m = hashlib.sha1()
         concat = ""
 
-        np.set_printoptions(precision=16)
```
loriab, line 771: Would the below do what you need or is the precision=16 actually throwing off your hashes in some way?

```
with np.printoptions(precision=16):
    for field
       ...
       concat +=
```

awvwgk: From my testing this actually does not change any hashes. It just is surprising to have `mol.get_hash()` change the output of printed arrays.

loriab (approved):

PR #318, opened 2023-08-03
Allow floating point numbers for multiplicities

### LINKED ISSUE #317: Floating point number allowed for molecular charge but not molecular multiplicity
**Describe the bug**

QCElemental allows floating point numbers for molecular charges, however when computing the spin with the fractional electron there is no way to represent the molecular multiplicity with just an integer.

**To Reproduce**

```python
import qcelemental as qcel
mol = qcel.models.Molecule(
    geometry=[0.0, 0.0, 0.5, 0.0, 0.0, -0.5],
    symbols=["H", "He"],
    molecular_charge=0.5,
    molecular_multiplicity=1.5,  # not possible
)
```

**Expected behavior**

Consistent behavior for specifying molecular charge and molecular multiplicity.

**Additional context**

awvwgk:
Closes #317 

## Description

- make behavior consistent between molecular / fragment charges and molecular / fragment multiplicities

## Changelog description

- allow floating point numbers for multiplicities

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcelemental/molparse/chgmult.py
+++ b/qcelemental/molparse/chgmult.py
@@ -101,6 +101,8 @@ def validate_and_fill_chgmult(
     ------
     qcelemental.ValidationError
         When no solution to input arguments subject to the constraints below can be found.
+    TypeError
+        When fractional multiplicity is provided.
 
     Notes
     -----
@@ -299,6 +301,19 @@ def validate_and_fill_chgmult(
     log_full = verbose >= 2
     log_brief = verbose >= 2  # TODO: Move back to 1
     text = []
+
+    def int_if_possible(val):
+        if isinstance(val, float) and val.is_integer():
+            return int(val)
+        else:
+            return val
+
+    molecular_multiplicity = int_if_possible(molecular_multiplicity)
+    fragment_multiplicities = [int_if_possible(m) for m in fragment_multiplicities]
+    if isinstance(molecular_multiplicity, float) or any(isinstance(m, float) for m in fragment_multiplicities):
+        raise TypeError(
+            f"validate_and_fill_chgmult() cannot handle fractional multiplicity. m: {molecular_multiplicity}, fm: {fragment_multiplicities}"
+        )
 
     felez = np.split(zeff, fragment_separators)
     nfr = len(felez)
--- a/qcelemental/models/molecule.py
+++ b/qcelemental/models/molecule.py
@@ -179,7 +179,7 @@ class Molecule(ProtoModel):
         description="Additional comments for this molecule. Intended for pure human/user consumption and clarity.",
     )
     molecular_charge: float = Field(0.0, description="The net electrostatic charge of the molecule.")  # type: ignore
-    molecular_multiplicity: Union[int, float] = Field(1, description="The total multiplicity of the molecule.")  # type: ignore
+    molecular_multiplicity: float = Field(1, description="The total multiplicity of the molecule.")  # type: ignore
 
     # Atom data
     masses_: Optional[Array[float]] = Field(  # type: ignore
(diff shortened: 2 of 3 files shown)


loriab (commented):
In discussions, @bennybp doesn't anticipate any trouble with the database. Thanks for ensuring existing molecule hashes won't change.

It just occurred to me that qcng harnesses could be collecting a non-meaningful float multiplicity that has heretofore gotten cast to int but will henceforth be float unless we do some processing to cast to int for whole numbers. A lot of molecules in the database could be hard to attach new calcs to. Is this a problem that needs a database migration, Ben?

loriab:
@awvwgk, your `Union[int, float]` construction looked right to me, but upon testing, it was casting values like 3.00001 to 3 in an unpredictable way. I switched the multiplicities to being `float` like charges are and added a pydantic validator to coerce them to ints where possible.

Note that to create a fractional multiplicity Mol, one has to pass `validate=False` to bypass the molparse chgmult physics logic. The qcel test suite passes cleanly, and part of psi4's test suite passes cleanly, so I'm reasonably convinced that these changes are harmless. Do they still do what you need?

awvwgk:
Why would you have an error for fractional multiplicities when you can pass fractional charges with validation error?

loriab:
> Why would you have an error for fractional multiplicities when you can pass fractional charges with validation error?

"without validation error"?

To be clear, mult=3.0 is fine; only values like 3.01 error. From what I recall, the chg/mult logic in molparse wasn't designed for non-int mult _or_ charge. But float charge passes through at least, whereas as-is it errors with float mult -- it would need some changes to accommodate float mult. So how deeply do you want/need fractional quantities incorporated? A small change could probably allow them if fully specified (i.e., no blanks in `fragment_multiplicities` if nfrag>1 and `molecular_multiplicity` is fractional).

qcelemental/molparse/chgmult.py
```diff
@@ -300,6 +302,19 @@ def validate_and_fill_chgmult(
...
+            return int(val)
+        else:
+            return val
+
+    molecular_multiplicity = int_if_possible(molecular_multiplicity)
+    fragment_multiplicities = [int_if_possible(m) for m in fragment_multiplicities]
+    if isinstance(molecular_multiplicity, float) or any(isinstance(m, float) for m in fragment_multiplicities):
+        raise TypeError(
+            f"validate_and_fill_chgmult() cannot handle fractional multiplicity. m: {molecular_multiplicity}, fm: {fragment_multiplicities}"
+        )
```
awvwgk, line 316: If this is not blocking the use of multiplicities like `1.5` I am fine with merging this change. I don't see an equivalent check for charges therefore I am a bit concerned whether this change only in principle allows fractional multiplicities.

awvwgk:
Lori @loriab, can we merge this PR or is anything blocking it?

awvwgk:
What is the status of this PR, is anything missing?

loriab:
Sorry for the delay. So long as what you need is unvalidated pass-through of non-int chg and mult, I think all this PR needs is an extra message for charge like the new warning one for mult in the case of non-integers. I'd better add some non-int charge tests, too. It's true that float charge has been there from the beginning but possibly never exercised.

awvwgk:
I think that works for now as a solution

loriab:
Ok, https://github.com/loriab/QCElemental/commit/342d7cc4adea5367d950dc04d2324946def009db is my current proposal for allowing float multiplicity. It's sitting atop #343 for now.

Investigation showed that the validation routine does handle float charges and checking internal consistency between total and fragment charges. It doesn't try to assign fractional multiplicities to go along with fractional charges, and this is probably edge-case enough in QC to not demand those checks. Accordingly, the validation routine has been made float-tolerant for multiplicities, so validate=T/F allowed. It doesn't try to do any fancy guessing, and you may find cases where its handling should be loosened, tightened, or better coordinated with charge in future.

PR #322, opened 2023-08-11
pyenv which hint

loriab:
## Description
See https://github.com/MolSSI/QCEngine/issues/292 pyenv returns a path from `shutil.which` that, if its an unactivated conda env under pyenv management, isn't actually ready to run, so `qcelemental.which` is deceptive. Apart from checking err better on the qcengine side, this PR better handles the situation.

## Status

- [x] Code base linted
- [x] Ready to go


loriab:
@coltonbh, if you have any suggestions for special handling in `which_import`, please let me know.

coltonbh:
Sure! The case you're handling is essentially that someone is using `pyenv`, which is pretty industry standard (at least outside academia) for installing all your python interpreters (i.e., python 3.10.8, 3.11.2, some miniconda install or anaconda install). 

If you run which `psi4` from the command line and you've used a `pyenv` installed miniconda/anaconda to install `psi4` (or any other executable), you'll see the following result. `which` will return the path to the executable, but when you try to run that executable you are given the following error by `pyenv`, essentially `pyenv` knows the executable is installed somewhere in one of your conda envs but you need to run `conda activate ...` first.

```sh
~ ❯ which psi4
/home/cbh/.pyenv/shims/psi4
~ ❯ /home/cbh/.pyenv/shims/psi4
pyenv: psi4: command not found

The `psi4' command exists in these Python versions:
  miniconda3-4.7.12/envs/psi4
  miniconda3-4.7.12/envs/psi4-test
  miniconda3-4.7.12/envs/psi4v1.6

Note: See 'pyenv help global' for tips on allowing both
      python2 and python3 to be found.

```

This code covers this case and provides a notification to end users. Though now that I'm looking at it I think the `{path.split('/')[-1]` line will only be correct if your environment is named the same as the executable. So you could just say `f"Please activate a conda environment that contains {program}."

```python
@lru_cache  # Improves perf of tests when repeatedly called
def prog_available(program: str) -> bool:
    """Check if a program is available on the system.

    Args:
        program: The program to check.

    Returns:
        True if the program is available, False otherwise.
    """
    path = shutil.which(program)
    if not path:
        return False
    # Perform secondary checks
    if ".pyenv/shims" in path:
        # Executable is in non activated pyenv environment; often a conda env
        print(
            f"Program {program} is in a pyenv shim. It may become available if you "
            f"run 'conda activate {path.split('/')[-1]}'"
        )
        return False
    return True
```

Any other questions lemme know :)

loriab:
Thanks for the background -- I've adapted the message accordingly. I think this takes care of `qcel.which` based on `shutil.which`. 

One thing I was curious about is if you had run into any pyenv curiosities with `qcel.which_import` based on `importlib.util.find_spec` such that we need to adapt it, too.

coltonbh:
None that I know right off hand. But the qcel import system was a little complex so that's part of why I have some of my own checkers lol ;P I couldn't always follow the plot fully.

loriab:
Thanks, gtk.

Yeah, sometimes one wants path/None, sometimes path/error, sometimes T/F. And it was more compact to complicate the qcel routines and standardize messages there than to repeat across all the qcng harnesses. I probably ought to add a table like the below to the which/which_import docstring -- I couldn't have written it from memory. The submodule bit doesn't help the which_import clarity :-)

```
+-------------+-------------+---------------------------------+---------------------------+
| return_bool | raise_error | action if found                 | action if not found       |
+-------------+-------------+---------------------------------+---------------------------+
| F (default) | F (default) | return <path to command> string | return None               |
+-------------+-------------+---------------------------------+---------------------------+
| T           | F (default) | return True                     | return False              |
+-------------+-------------+---------------------------------+---------------------------+
| F (default) | T           | return <path to command> string | raise ModuleNotFoundError |
+-------------+-------------+---------------------------------+---------------------------+
| T           | T           | return True                     | raise ModuleNotFoundError |
+-------------+-------------+---------------------------------+---------------------------+
```

Lnaden (approved):

Issue #324, opened 2023-09-19
Floating-point precision error in calculating angles


philipmnel:
**Describe the bug**
When using `qcel.util.measure_coordinates` to measure an angle between coordinates, an angle of 180° can cause a `nan` value to be returned. The bug is caused by a floating-point precision error in calculating the cosine of the angle; if the cosine is greater than `1.0` due to noise then `np.arccos` will return `nan`.

**To Reproduce**
```
import qcelemental as qcel
import numpy as np

geometry= np.asarray([[0, 0, 0], [1, 0, 0], [2, 0, 0]])
angle = qcel.util.measure_coordinates(geometry, [0,1,2], True)
print(angle)

geometry= np.asarray([[-5.53102237, -12.51893475, -8.92439275], 
                      [-5.53102237, 2.9394029, 1.97762624], 
                      [-5.53102237, 18.39774056, 12.87964524]])
angle = qcel.util.measure_coordinates(geometry, [0,1,2], True)
print(angle)
```
The first prints `180.0` as expected, the second returns a `nan` (calculated cosine is `1.0000000000000002`). 

Using qcelemental 0.25.0

**Expected behavior**
The calculated cosine should be properly bounded within [-1, 1].

**Additional context**

loriab:
closed by #325

PR #325, opened 2023-09-19
Fix floating-point bug in `compute_angle`

philipmnel:
## Description
Fixes the bug described in #324 by ensuring that the calculated cosine has a valid value between [-1, 1].

## Changelog description
Fixed a bug causing `qcel.util.measure_coordinates` to return `nan` values.

## Status

- [ ] Code base linted
- [x] Ready to go


loriab (approved):
lgtm, thanks!

np.clip was introduced in v1.17 in 2019, which seems reasonable.

PR #326, opened 2023-09-29
Add eigenvalues_and_occupations to wavefunction protocols

bennybp:
## Description
Adds the ability to save only eigenvalues and occupations. Tests have been added.

Background: We need to obtain/store orbital energies for millions of molecules, and we can't afford to also store orbitals for all of them :)

No need to modify QCEngine as the stripping of fields is done in QCElemental.

## Changelog description
Add `eigenvalues_and_occupations` as an option to WavefunctionProtocols

## Status

- [X] Code base linted
- [X] Ready to go


loriab (approved):
lgtm!

PR #338, opened 2024-06-18
adapt for numpy v2

loriab:
- [x] Ok, this fixes up qcel for numpy v2. But poetry won't actually let me specify the loosest constraints (use pint v0.24 if numpy v2), so pyproject.toml a little more restrictive than reality.
- [x] ~also adds chg/mult to Molecule repr -- that's for eyes so should be harmless.~ moved to #340

## Status

- [x] Code base linted
- [x] Ready to go


Lnaden (approved):
Looks good to me!

PR #343, opened 2024-07-26
add hashes and rearrange mult handling

loriab:
## Description

This PR is preparatory to #318.

- [x] NEW: adds a check to the basic Molecule pydantic model so multiplicities <1 get caught. This route is accessed when `Molecule(..., validate=False)`. I believe this is a universally good check, but it is a new restriction. Any weird positron, fractional electron, etc. reason to not catch less-than-one mult?
- [x] add many hash checks (incl. all the chgmult variations in `validate_and_fill_chgmult` testing) to molecules, so we'll know if code starts changing the hash and thereby disrupting the QCA server.
- [x] `qcel.molparse` has a different error message for less-than-one `fragment_multiplicities` you try to put through it. `qcel.molparse` is accessed whenever `Molecule(...)` or `Molecule(..., validate=True)` is called.
- [x] `qcel.molparse` newly allows floats that are ints (e.g., `1.0`) for multiplicity. previously it would raise an error about not being an int.
- [x] `qcel.molparse` newly has a dedicated error for any less-than-one `molecular_multiplicity` you try to put through it. Previously it raised the usual "Inconsistent" error.
- [x] more tests that chgmult are processing as expected, including some with fractional _charges_ (https://github.com/MolSSI/QCElemental/pull/343/files#diff-29ef453faf072f4ee853122f0f64f6ba594c00dc44599a1f5138946619d9951fR90-R96) for the first time.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcelemental/models/molecule.py
+++ b/qcelemental/models/molecule.py
@@ -419,6 +419,8 @@ class Molecule(ProtoModel):
                 raise ValueError(
                     "Fragment Multiplicities must be same number of entries as Fragments"
                 )
+        if any([m<1.0 for m in v]):
+            raise ValueError(f"Fragment Multiplicity must be positive: {v}")
         return v
 
     @validator("molecular_multiplicity")
--- a/qcelemental/models/molecule.py
+++ b/qcelemental/models/molecule.py
@@ -401,14 +401,32 @@ class Molecule(ProtoModel):
             v = np.array([True for _ in range(n)])
         return v
 
-    @validator("fragment_charges_", "fragment_multiplicities_")
+    @validator("fragment_charges_")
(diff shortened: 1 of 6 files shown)


loriab:
@awvwgk , is enforcing a lower bound of 1.0 on multiplicity a sensible thing to do even for fractional electron, or is it restricting?

awvwgk:
I believe a lower bound of 1.0 is sensible, I can't think of a usecase even for a fractional electron which could not be represented this way.

Derk @derkkooi, do you know any case that is not representable with this lower bound of multiplicity?

bennybp (approved):

PR #346, opened 2024-09-05
Csse pyd2 505

loriab:
## Description

- [x] add a pair of fixtures so that tests can grab models from v1 or v2 (right now v1) locations in the code to effectively parameterize over the two versions. the `Molecule` fixture could be done through the `schema_versions` one, but this saves some code since many tests only use Molecule
- the test_model_results has a cascade of fixtures that the new schema_versions _seems_ to be operating correctly with without much thought on my part, but I'm keeping an eye on it when v1 and v2 are actually incompatible
- [x] deleted file is duplicate of one moved to tests/ last year

## Status

- [x] Code base linted
- [x] Ready to go


loriab:

krachwal (approved):

davidbrownell (approved):

PR #347, opened 2024-09-09
Csse pyd2 510

loriab:
## Description

- [x] Provide a version of QCSchema expressed in Pydantic v2 API lodged in `qcelemental.models.v2`. It is not accessed by default.
- [x] First three commits are copying in a set of `models.v1` into `models`, applying the `models` portion of @Lnaden 's #321, and copying the result down to `models.v2`. Only the 4th commit d515a5f is changes to make it all work.
- [x] The outlandish pyproject.toml changes are because to build docs for pyd v2 models one needs a recent sphinx, autodoc-pydantic, numpydoc, etc., but poetry insists it be solvable for py37.
- [x] Ended up pulling in #342 to fix and/or see the true errors in docs build.
- [x] It's awkward to deprecate in code, so this is official notice that homegrown AutoDoc will not be in qcdk v2 and could be removed from v1. All of QCArchive uses autodoc-pydantic instead, so this is notice to any downstream models. The autodoc changes from #321 will be discarded (sorry, Levi, that was a lot of work iirc.

## Status

- [x] Code base linted
- [x] Ready to go
- REMINDER, DON'T SQUASH MERGE THIS ONE


loriab:

krachwal (approved):

PR #348, opened 2024-09-10
Csse pyd2 510 Part 2

loriab:
## Description

* ~currently atop #347~
- [x] pick over remaining relevant parts of @Lnaden #321 (below), then patch 'em up. so first commit is from #321 (note not a literal cherry-pick like #347) and second commit a1f7dcc are the fixes to get it working.
- [x] Fix up `DFTFunctional`, `CPUInfo` to be pydantic v2 API only. (afaik, this was an effort that didn't go anywhere. they're not used in qcng. Anyone know of usage in the wild downstream?)
- [x] Fix up `Datum` to be pydantic v2 API only. This is well-used, esp for NIST physical constant, but afaik it's never within a QCSchema model, so safe to support only one version.
- [x] Fix up serialization of models. Tentatively, trying an in-class try/except/try/except to handle serializing both pyd v1 and v2 API models. It works for tests, but we'll see about real life.

## Status

- [x] Code base linted
- [x] Ready to go
- DON'T SQUASH


loriab:

krachwal (approved):

PR #349, opened 2024-09-17
Csse pyd2 shim classes

loriab:
## Changelog description

- [x] Add dummy classes and warnings so that e.g., `from qcelemental.models.procedures import OptimizationInput` can still work.

## Status

- [x] Code base linted
- [x] Ready to go

PR #352, opened 2024-09-24
Csse pyd2 model convert classes

loriab:
## Description

~currently atop #349~
- [x] main Input/Result classes and FailedOperation gained a `convert_v(<ver>)` function to transform to other version. transformation trivial at this point b/c this is pydantic's purpose, but functions will differentiate at the rearrangement milestone. Needed to actually validate `schema_version` field so it would overwrite at conversion.
- [x] added `model_dump`, `model_dump_json`, `copy_model` to v1 so easier to write code that can take both model types. Levi had done the reverse and added `dict`, `json`, `copy` to v2, but I'm thinking this is the better way. Haven't removed Levi's fns.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -35,8 +35,9 @@ New Features
 
 Enhancements
 ++++++++++++
+* ``v2.FailedOperation`` field `id` is becoming `Optional[str]` instead of plain `str` so that the default validates.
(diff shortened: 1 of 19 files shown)


loriab:

krachwal (approved):

PR #357, opened 2024-11-05
Csse layout 550 error

loriab:
## Description

- [x] rm `v2.<>Result.error` field and forgive its empty presence in `v1.<>Result.convert_v()` fn
- [x] enforce success=T/F in `v2.<>Result` and `v2.FailedOperation`
- [x] changelog for details
- py37 is on the chopping block
- also, the `schema_version` stamping instead of popping the version was convenient for the `convert_v` function when v1 and v2 were same layout. but the pop strategy might be better long-haul.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcelemental/models/v2/procedures.py
+++ b/qcelemental/models/v2/procedures.py
@@ -350,6 +350,9 @@ class TorsionDriveResult(TorsionDriveInput):
 
         dself = self.model_dump()
         if version == 1:
+            if dself["optimization_spec"].pop("extras", None):
+                pass
+
             self_vN = qcel.models.v1.TorsionDriveResult(**dself)
 
         return self_vN


loriab:

qcelemental/models/v1/procedures.py
```diff
@@ -332,8 +352,18 @@ def convert_v(
         if check_convertible_version(version, error="TorsionDriveResult") == "self":
             return self
 
+        opthist_class = next(iter(self.optimization_history.values()))[0].__class__
         dself = self.dict()
         if version == 2:
+            # remove harmless empty error field that v2 won't accept. if populated, pydantic will catch it.
+            if dself.pop("error", None):
```
davidbrownell, line 359: Consider removing the if statement (since it doesn't do anything) and change to `dself.pop("error", None)`.

davidbrownell (approved):

qcelemental/models/v1/procedures.py
loriab: fixed here and two other spots in next PR

PR #358, opened 2024-11-12
Csse layout 536 input_data

loriab:
~currently atop #357~ 

## Description

* see changelog
* `AtomicResult(AtomicInput)` --> `AtomicResult()` with `AtomicResult.input_data = AtomicInput`
* counterpart to https://github.com/MolSSI/QCEngine/pull/459

## Status

- [x] Code base linted
- [x] Ready to go

PR #359, opened 2024-11-20
Csse layout specfication

loriab:
~currently atop #358~ 

## Description

- [x] see changelog
- [x] this moves many of the fields in `AtomicInput` into a submodel `AtomicSpecification` that can be shared with other input schema

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcelemental/models/v1/results.py
+++ b/qcelemental/models/v1/results.py
@@ -628,7 +628,15 @@ class AtomicInput(ProtoModel):
             return self
 
         dself = self.dict()
+        spec = {}
         if version == 2:
+            dself.pop("schema_name")  # changes in v2
+
+            spec["driver"] = dself.pop("driver")
+            spec["model"] = dself.pop("model")
+            spec["keywords"] = dself.pop("keywords", None)
+            spec["protocols"] = dself.pop("protocols", None)
+            dself["specification"] = spec
             self_vN = qcel.models.v2.AtomicInput(**dself)
 
         return self_vN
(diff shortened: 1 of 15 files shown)


loriab:

qcelemental/models/v1/results.py
```diff
@@ -619,16 +613,25 @@ def _version_stamp(cls, v):
...
         """Convert to instance of particular QCSchema version."""
         import qcelemental as qcel
 
-        if check_convertible_version(version, error="AtomicInput") == "self":
+        if check_convertible_version(target_version, error="AtomicInput") == "self":
             return self
 
         dself = self.dict()
-        if version == 2:
+        spec = {}
```
davidbrownell, line 625: Consider moving this declaration into the if clause.

qcelemental/models/v2/results.py
```diff
@@ -653,6 +647,52 @@ class AtomicResultProtocols(ProtoModel):
...
+        dself = self.model_dump()
+        if version == 1:
+            loss_store["protocols"] = dself.pop("protocols")
+            loss_store["program"] = dself.pop("program")
+
+            if loss_store:
+                dself["extras"]["_qcsk_conversion_loss"] = loss_store
+
+            self_vN = qcel.models.v1.QCInputSpecification(**dself)
+
```
davidbrownell, line 692: If I understand this logic correctly, it seems like `check_convertible_version` will return `self` when `version` == 2. Is that correct?

To help a reader be more confident in their assumptions around the code, consider adding an else clause that asserts false, just to show that it is never an expected scenario.

```
else:
    assert False, version
```

qcelemental/models/v2/results.py
```diff
@@ -708,16 +743,25 @@ def _version_stamp(cls, v):
...
+        if target_version == 1:
+            dself["driver"] = dself["specification"].pop("driver")
+            dself["model"] = dself["specification"].pop("model")
+            dself["keywords"] = dself["specification"].pop("keywords", None)
+            dself["protocols"] = dself["specification"].pop("protocols", None)
+            dself["extras"] = dself["specification"].pop("extras", {})
+            dself["specification"].pop("program", None)  # TODO store?
+            assert not dself["specification"], dself["specification"]
+            dself.pop("specification")  # now empty
+
```
davidbrownell, line 764: Consider adding an else clause similar to what is described above.

qcelemental/models/v2/results.py
```diff
@@ -910,22 +954,24 @@ def _native_file_protocol(cls, value, info):
...
-            input_data = dself.pop("input_data")
+        if target_version == 1:
+            # for input_data, work from model, not dict, to use convert_v
+            dself.pop("input_data")
+            input_data = self.input_data.convert_v(1).model_dump()  # exclude_unset=True, exclude_none=True
             input_data.pop("molecule", None)  # discard
             input_data.pop("provenance", None)  # discard
             dself["extras"] = {**input_data.pop("extras", {}), **dself.pop("extras", {})}  # merge
             dself = {**input_data, **dself}
+
```
davidbrownell, line 974: Consider adding an else clause here too.

davidbrownell (approved):

qcelemental/models/v2/results.py
loriab: Yes, `check_convertible_version` is a weird ~return of `True|"self"|raised error`. I'll add the false clause you mentioned in the next PR, as it should go in several classes.

PR #361, opened 2024-11-26
poetry to setuptools; min py38

loriab:
## Description

- [x] After all-but-py37 builds spontaneously broke in the last few days, it's easier to move the build backend back in line with other QCA projects (QCFractal and QCManyBody) rather than find the issue with poetry's solve-all-at-once and highly-version-specified format.
- [x] This should hit most poetry uses and examples. I've already confirmed that new builds on `master` didn't break, so fine targeting this to `next2024`.
- [x] Dialed back the pydantic versions for docs build. There's some errors with latest pydantic, so that pin needs investigating and releasing.
- [x] Also bump the min python to 3.8+. Even that is end-of-life but QCEngine test lanes still use it.
- There will likely be refinements and corrections to this switchover, but this is a largely working start.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -31,7 +31,7 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v3
       - name: Install QCElemental (full deps)
-        if: (!(matrix.python-version == '3.8' || matrix.python-version == '3.10'))
+        if: matrix.python-version != '3.8' && matrix.python-version != '3.10'
         run: pip install '.[test,viz,align,standard]'
       - name: Install QCElemental (min deps)
         if: matrix.python-version == '3.8' || matrix.python-version == '3.10'
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -31,10 +31,10 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v3
       - name: Install repo with poetry (full deps)
-        if: matrix.python-version != '3.9'
+        if: (!(matrix.python-version == '3.8' || matrix.python-version == '3.10'))
         run: pip install '.[test,viz,align]'
       - name: Install repo with poetry (min deps)
-        if: matrix.python-version == '3.9'
+        if: matrix.python-version == '3.8' || matrix.python-version == '3.10'
         run: pip install '.[test]'
       - name: Run tests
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -1,126 +1,98 @@
...
-    { version = "^1.8.0", python = ">=3.9,<4.0.0" }
+test = [
+    "pytest",
+    "pytest-cov",
+    "codecov",
 ]
-docutils = [
-    { version = "<0.19", python = "<3.9" },
-    { version = "0.20.1", python = ">=3.9,<4.0.0" }
+dev = [
```
loriab, line 59:
```suggestion
lint = [
```

loriab:

.github/workflows/CI.yaml
```diff
@@ -30,16 +30,14 @@ jobs:
...
-      - name: Install poetry
-        run: pip install poetry
-      - name: Install repo with poetry (full deps)
-        if: matrix.python-version != '3.9'
-        run: poetry install --no-interaction --no-ansi --all-extras
-      - name: Install repo with poetry (min deps)
-        if: matrix.python-version == '3.9'
-        run: poetry install --no-interaction --no-ansi --extras test
+      - name: Install QCElemental (full deps)
+        if: (!(matrix.python-version == '3.8' || matrix.python-version == '3.10'))
```
davidbrownell, line 34: Nit: I had to spend more time than I'd care to admin looking at this step and the next one to figure out why they both exist (my mind was thinking that they both had the same conditional).

I don't think that this needs a change, but some options that may have saved time for a person who thinks like me :) :

*   Comments to indicate that one clause was for 3.8 and 3.10 and another for other python versions
*   Convert `!(matrix.python-version == '3.8' || matrix.python-version == '3.10')` to `matrix.python-version != '3.8' && matrix.python-version != '3.10'`
*   Convert `!(matrix.python-version == '3.8' || matrix.python-version == '3.10')` to `matrix.python-version == '3.9' || matrix.python-version == '3.11' || matrix.python-version == '3.12' || matrix.python-version == '3.13'` (I don't like that this option needs to be maintained as python versions come and go)

davidbrownell (approved):

.github/workflows/CI.yaml
loriab, line 34:
```suggestion
        if: matrix.python-version != '3.8' && matrix.python-version != '3.10'
```

PR #376, opened 2025-10-01
fix: if verbose, do log info statements

steph409:
## Description

there is a small issue in the molecule. a higher verbosity, in general, results in more log messages. however, in the molecule-model, it is implemented the wrong way around. a higher verbosity results in quiet=True, which is not expected.

## Changelog description

fix verbosity of models/molecule.py log messages

## Status

- [x] Code base linted
- [x] Ready to go


loriab (approved):
Thanks for the fix — that indeed looks to have been bad logic.

PR #379, opened 2025-10-07
avoid uncontrollable printing

### LINKED ISSUE #368: Unavoidable print output.
**Describe the bug**
When making a molecule, in line 522 of `QCElemental.molparse.chgmult.py` in the function `validate_and_fill_chgmult` there is a sub function `reconcile` with a printout:
```
if verbose > -1:
            print("\n\n" + "\n".join(text))
```
It seems like this should be guarded by `log_full` or something. 

**Expected behavior**
Controllable print out.

loriab:
## Description

Addresses #368 as stated. I moved the printing to the error msg and allowed verbose to be passed through from Mol creation. But I wonder if your trouble was definitely around line 522 (which only printed when it was about to throw) or similar content around line 563, @jaclark5?

## Status

- [x] Code base linted
- [x] Ready to go

closes #368

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -20,8 +20,13 @@ Changelog
 .. +++++
 
 
-0.30.0 / 2025-MM-DD (Unreleased)
+0.30.0 / 2026-01-07
 -------------------
+
+Enhancements
+++++++++++++
+- (:issue:`368`, :pr:`379`) Remove uneeded printing and hook up verbosity control for
+  charge/multiplicity reconciliation for ``Molecule.from_data``. @jaclark5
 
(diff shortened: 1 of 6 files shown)


jaclark5:
@loriab I made this issue a while ago so I don't remember for sure, but I did find this annoying and did my best to run it down so if I trust "past-me" then I'll say I'm 97% sure this is the line that's an issue.

I added a reproducible example in my original issue, https://github.com/MolSSI/QCElemental/issues/368#issuecomment-3492466706, to test it.

jaclark5:
@loriab I can confirm that this branch resolves the issue

PR #388, opened 2026-02-15
Some small fixes for v2 Molecule

bennybp:
## Description

3 small fixes for the v2  molecule in next2026. No change in behavior or interface

1. Need to import sys (and cleans up other unused imports)
2. Fixes printing for float multiplicity
3. Utility should use `type(self)` rather than `Molecule` in order to handle being called from derived classes
4. Remove unused NDArray stuff for python < 3.9
5. Fix typing of function parameters & return types
6. Improve logic of molecule `@property` (connectivity doesn't need an aliased attribute, and don't use `__dict__` when aliases are used)

## Changelog description

Fixes imports, printing, and returned objects in v2 `Molecule` class

## Status

- [X] Code base linted
- [X] Ready to go

### DIFF
--- a/qcelemental/models/v2/molecule.py
+++ b/qcelemental/models/v2/molecule.py
@@ -5,13 +5,14 @@ Molecule Object Model
 import collections
 import hashlib
 import json
+import sys
 import warnings
 from functools import partial
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Literal, Optional, Tuple, Union, cast
 
 import numpy as np
-from pydantic import Field, constr, field_validator, model_serializer
+from pydantic import Field, field_validator, model_serializer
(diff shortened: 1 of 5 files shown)


qcelemental/models/v2/molecule.py
```diff
@@ -2,24 +2,21 @@
...
 import numpy as np
-from pydantic import Field, constr, field_validator, model_serializer
+from pydantic import Field, field_validator, model_serializer
 from typing_extensions import Annotated
 
-try:
-    import nglview
-except ModuleNotFoundError:
-    # import is purely for forward reference for docs-build. import is not required except for Molecule.show()
-    pass
```
loriab, line 21: I'll probably try putting this back b/c there are more docs warnings with this missing.

loriab (approved):

PR #390, opened 2026-02-17
Suppress warning and fix docs warnings for 0.50.0rc2

loriab:
## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -211,7 +211,7 @@ intersphinx_mapping = {
     "numpy": ("https://numpy.org/doc/stable/", None),
     "scipy": ("https://docs.scipy.org/doc/scipy/", None),
     "matplotlib": ("https://matplotlib.org/stable/", None),
-    "qcengine": ("https://molssi.github.io/QCEngine/", None),
+    "qcengine": ("https://molssi.github.io/QCEngine/next", None),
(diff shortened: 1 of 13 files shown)


bennybp:
I think the dependencies for docs can be cleaned up (which might also help some of these warnings). For example, pydantic is in the core dependencies, and `pydantic-settings` isn't ever used.

Autodoc supports type-hints (I believe) so we could possibly do away with `sphinx-autodoc-typehints` dependency and in docs `conf.py` unless we have some specific reason

docs/conf.py
```diff
@@ -200,6 +199,7 @@ def setup(app):
 extlinks = {
     "issue": ("https://github.com/MolSSI/QCElemental/issues/%s", "GH#%s"),
     "pr": ("https://github.com/MolSSI/QCElemental/pull/%s", "GH#%s"),
+    "docs": ("https://MolSSI.github.io/QCElemental/%s", "Docs %s"),
```
loriab, line 202: alternative 

```
`Docs for QCSchema v1<https://MolSSI.github.io/QCElemental/v0.30.2/>`_
```
seems to process address as lowercase link that fails to resolve so defining this shortcut

bennybp (approved):

PR #391, opened 2026-03-06
Some cleanup and further improvements

bennybp:
## Description

A bit more cleanup and stabilizing before making backwards-incompatible changes. No tests are changed, and user-facing behavior should be the same.

* Remove unused config options (`canonical_repr` and `repr_style`)
* Remove unused `serialize_default_excludes` functionality
* Add convert_v to some models (`AtomicProperties`, `AtomicProtocols`, `WavefunctionProperties`).
  * Right now, `schema_name` is automatically excluded because it is the default value. But that will change
* Make Molecule and `WavefunctionProperties` tolerant of `None` being passed in during construction
  * Will be more important when various models will have `None` values with `model_dump`
* Clean up other unused code

## Status

- [X] Code base linted
- [X] Ready to go


qcelemental/models/v2/atomic.py
```diff
@@ -307,6 +307,28 @@ def dict(self, *args, **kwargs):
...
+            dself.pop("schema_name", None)
+
+            if target_version == 1:
+                self_vN = qcel.models.v1.AtomicResultProperties(**dself)
+            elif target_version == QCEL_V1V2_SHIM_CODE:
+                self_vN = qcel.models._v1v2.AtomicProperties(**dself)
+        else:
+            assert False, target_version
+
+        return self_vN
```
loriab, line 330: Doesn't this need a counterpart in models/v1?

Uncomment https://github.com/MolSSI/QCElemental/blob/next2026/qcelemental/tests/test_model_results.py#L1272 for any models with new convert_v

qcelemental/models/v2/atomic.py
```diff
@@ -671,6 +704,28 @@ class AtomicProtocols(ProtoModel):
 
     model_config = ExtendedConfigDict(force_skip_defaults=True)
 
+    def convert_v(
+        self, target_version: int, /
+    ) -> Union["qcelemental.models.v1.AtomicProtocols", "qcelemental.models.v2.AtomicProtocols"]:
```
loriab:
```suggestion
    ) -> Union["qcelemental.models.v1.AtomicResultProtocols", "qcelemental.models.v2.AtomicProtocols"]:
```

ditto the v1 fn and the test

qcelemental/models/v2/common_models.py
```diff
@@ -46,7 +43,7 @@ class Model(ProtoModel):
     )
 
     # basis_spec: BasisSpec = None  # This should be exclusive with basis, but for now will be omitted
-    model_config = ProtoModel._merge_config_with(canonical_repr=True, extra="allow")
+    model_config = ProtoModel._merge_config_with(extra="allow")
```
loriab, line 46: Perhaps we need a `test_model_survey_extrafields` to add to the test_model_survey_* suite to check for extra=allow/forbid/ignore . I checked Specification the other day, and they were forbid. Provenance I know is allow. Hard to keep track and enforce w/o a test.

loriab (commented):
Serialization changes look reasonable, but I'm more believing than reviewing on those.

qcelemental/models/v2/common_models.py
bennybp: I started working on one, but we don't have the models in the testing fixtures (?). From what I can see, only Model and Provenance allow. By default, all is forbidden by default in Protomodel

qcelemental/models/v2/molecule.py
```diff
@@ -392,10 +385,7 @@ def __init__(self, orient: bool = False, validate: Optional[bool] = None, **kwar
 
         if validate:
             # Title case for consistency
-            if np.lib.NumpyVersion(np.__version__) >= "2.0.0b1":
-                values["symbols"] = np.char.chararray.title(self.symbols)
-            else:
-                values["symbols"] = np.core.defchararray.title(self.symbols)
+            values["symbols"] = np.char.chararray.title(self.symbols)
```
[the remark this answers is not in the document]
loriab: I've tested this on np v1 and v2. this'd solve the problem w/o restricting np

qcelemental/models/v1/results.py
```diff
@@ -300,6 +301,24 @@ def dict(self, *args, **kwargs):
...
+            return self
+
+        dself = self.model_dump()
+        if target_version in [2, QCEL_V1V2_SHIM_CODE]:
+            if target_version == 2:
+                self_vN = qcel.models.v2.AtomicProperties(**dself)
+        else:
+            assert False, target_version
+
+        return self_vN
```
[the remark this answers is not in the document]
loriab: yeah, I agree, SHIM_CODE needs to be removed from 314, too, right? you control the very limited body of code (psi4, qcng, qcf) that might pass `-12`, so that code should be restricting this route

qcelemental/models/v2/common_models.py
loriab: adding in 392

loriab (approved):

PR #393, opened 2026-03-11
Rework serialization and model dumping for v2 models

bennybp:
## Description

This reworks how qcelemental v2 models are serialized/dumped to bring them more inline with standard pydantic behavior.

1. Previously, models had custom options that forced unset fields to be excluded. This was generally ok, but also counterintuitive when some applications did want those default fields for compatibility/provenance reasons. The led to needing custom serializers.

2. Models had some (buggy) logic for handling serialization to/from aliased fields. This is now handled by pydantic v2 natively. (see model_config options `serialize_by_alias` and `validate_by_alias`).

3. Because of the above, models were not tolerant of `None` being passed in (`Molecule` and `WavefunctionProperties`), even though it was the default value. For example, constructing a Molecule with `fragments=None` would result in an error.

So this PR removes the customization, makes the models tolerant of `None`, and updates some of the tests.

**The somewhat breaking change:** by default now, calling `.model_dump()` will dump all fields, even unset and `None` fields. This can be controlled with `exclude_unset=True` and `exclude_none=True` (see  [pydantic docs](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_dump)). This is also true for things like `model_dump_json`, and pydantic functions like `to_jsonable_python`).

The `dict()` function is removed to also bring our terminology in line with pydantic.

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcelemental/models/v2/atomic.py
+++ b/qcelemental/models/v2/atomic.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from enum import Enum
 from functools import partial
 from typing import TYPE_CHECKING, Any, Dict, Literal, Optional, Set, Union
@@ -7,7 +9,7 @@ from pydantic import Field, field_validator
 
 from ...models import QCEL_V1V2_SHIM_CODE
 from ...util import provenance_stamp
-from .basemodels import ExtendedConfigDict, ProtoModel, check_convertible_version, qcschema_draft
+from .basemodels import ProtoModel, check_convertible_version, qcschema_draft
 from .basis_set import BasisSet
 from .common_models import DriverEnum, Model, Provenance
 from .molecule import Molecule
@@ -249,10 +251,8 @@ class AtomicProperties(ProtoModel):
         None, description="The number of CCSDTQ iterations taken before convergence."
     )
 
-    model_config = ProtoModel._merge_config_with(force_skip_defaults=True)
-
     def __repr_args__(self) -> "ReprArgs":
-        return [(k, v) for k, v in self.dict().items()]
+        return [(k, v) for k, v in self.model_dump(exclude_unset=True).items()]
 
     @field_validator(
         "scf_dipole_moment",
@@ -299,13 +299,6 @@ class AtomicProperties(ProtoModel):
         except (ValueError, AttributeError):
             raise ValueError(f"Derivative must be castable to shape {shape}!")
         return v
-
-    def dict(self, *args, **kwargs):
-        # pure-json dict repr for QCFractal compliance, see https://github.com/MolSSI/QCFractal/issues/579
-        # Sep 2021: commenting below for now to allow recomposing AtomicResult.properties for qcdb.
-        #   This will break QCFractal tests for now, but future qcf will be ok with it.
-        # kwargs["encoding"] = "json"
-        return super().model_dump(*args, **kwargs)
 
     def convert_v(
         self, target_version: int, /
@@ -536,11 +529,6 @@ class WavefunctionProperties(ProtoModel):
         None, description="Index to the beta-spin orbital occupations of the primary return."
     )
 
-    # Note that serializing WfnProp skips unset fields (and indeed the validator will error upon None values)
-    #   while including all fields for the submodel BasisSet. This is the right behavior, imo, but note that
-    #   v1 skips unset fields in BasisSet as well as the top-level model.
-    model_config = ProtoModel._merge_config_with(force_skip_defaults=True)
-
     @field_validator("scf_eigenvalues_a", "scf_eigenvalues_b", "scf_occupations_a", "scf_occupations_b")
     @classmethod
     def _assert1d(cls, v):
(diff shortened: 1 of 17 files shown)


qcelemental/models/v2/atomic.py
```diff
@@ -300,13 +300,6 @@ def _validate_derivs(cls, v, info):
             raise ValueError(f"Derivative must be castable to shape {shape}!")
         return v
 
-    def dict(self, *args, **kwargs):
-        # pure-json dict repr for QCFractal compliance, see https://github.com/MolSSI/QCFractal/issues/579
-        # Sep 2021: commenting below for now to allow recomposing AtomicResult.properties for qcdb.
-        #   This will break QCFractal tests for now, but future qcf will be ok with it.
-        # kwargs["encoding"] = "json"
-        return super().model_dump(*args, **kwargs)
```
loriab, line 308: I understand the wish, but I don't think you can just delete `dict()`. My reasoning is that for anyone starting to migrate from v1 to v2, w/o this the first thing they're going to hit is dozens of dicts everywhere, probably in several projects and in user scripts too. Even if they have dev/editable versions of everything and can change them, then they've done a lot of busy work and little migration work, they're left in a ragged state, and will need releases of everything to get operational again. First migration step should be the real stuff, not dict -> model_dump.

Idk if that means 
* (a) putting this function back (it's got a DeprecationWarning in at least one form)
* (b) replacing return model_dump with a message saying "uncomment me here" to postpone dict's retirement
* (c) making the fn available with an envvar

qcelemental/tests/test_molecule.py
```diff
@@ -473,14 +473,17 @@ def test_molecule_errors_shape(Molecule, water_molecule_data):
...
 
     assert isinstance(water_dimer_minima.model_dump_json(), str)
 
     with warnings.catch_warnings():
         warnings.simplefilter("ignore")
-        assert isinstance(water_dimer_minima.dict(encoding="json")["geometry"], list)
+        if "v2" in request.node.name:
+            assert isinstance(water_dimer_minima.model_dump(mode="json")["geometry"], list)
+        else:
+            assert isinstance(water_dimer_minima.dict(encoding="json")["geometry"], list)
```
loriab: might want to add some of these changes to the MIGRATION.rst guide

qcelemental/tests/test_molecule.py
```diff
@@ -208,7 +208,7 @@ def test_water_minima_data(Molecule, water_dimer_minima_data):
     # Give it a name
     with warnings.catch_warnings():
         warnings.simplefilter("ignore")
-        mol_dict = water_dimer_minima.dict()
+        mol_dict = water_dimer_minima.model_dump()
```
loriab: depending on final disposition of `def dict`, may want to add this back with a comment to leave alone for testing dict() operational.

qcelemental/tests/test_model_general.py
```diff
@@ -13,7 +13,10 @@ def test_result_properties_default_skip(request, schema_versions):
 
     assert pytest.approx(obj.scf_one_electron_energy) == -5.0
 
-    assert obj.dict().keys() == {"scf_one_electron_energy"}
+    if "v2" in request.node.name:
+        assert obj.model_dump(exclude_unset=True).keys() == {"scf_one_electron_energy"}
+    else:
+        assert obj.dict().keys() == {"scf_one_electron_energy"}
```
loriab: revisit when AtomicProperties None's decided

qcelemental/tests/test_model_results.py
```diff
@@ -812,7 +816,7 @@ def test_result_derivatives_array(request, schema_versions):
     assert obj.calcinfo_natom == 4
     assert obj.return_gradient.shape == (4, 3)
     assert obj.scf_total_hessian.shape == (12, 12)
-    assert obj.model_dump().keys() == {"calcinfo_natom", "return_gradient", "scf_total_hessian"}
+    assert obj.model_dump(exclude_unset=True).keys() == {"calcinfo_natom", "return_gradient", "scf_total_hessian"}
```
loriab: these three also revisit after AtomicProperties and WavefunctionProperties

qcelemental/tests/test_model_results.py
```diff
@@ -1335,7 +1339,12 @@ def test_model_survey_dictable(smodel1, smodel2, every_model_fixture, request, s
     instance = model(**data)
     with warnings.catch_warnings():
         warnings.simplefilter("ignore")
-        instance = model(**instance.dict())
+
+        if "v2" in anskey:
+            instance = model(**instance.model_dump())
+        else:
+            instance = model(**instance.dict())
```
loriab: revisit after def dict()

qcelemental/models/v2/molecule.py
```diff
@@ -614,27 +616,6 @@ def __eq__(self, other):
 
         return self.get_hash() == other.get_hash()
 
-    # UNCOMMENT IF NEEDED FOR UPGRADE REDO??
-    def dict(self, **kwargs):
-        warnings.warn("The `dict` method is deprecated; use `model_dump` instead.", DeprecationWarning)
-        return self.model_dump(**kwargs)
```
loriab, line 620: revisit after def dict() decided

qcelemental/models/v2/molecule.py
```diff
@@ -614,27 +616,6 @@ def __eq__(self, other):
...
-    def _serialize_molecule(self, handler) -> Dict[str, Any]:
-        default_result = handler(self)
-        output_dict = {}
-        for key, value in default_result.items():
-            # Could do this as a single comprehension dict, but this is easier to read
-            # Handle exclude unset is always true
-            if key not in self.model_fields_set:
-                continue
-            # Handle "by_alias" is always true
-            alias = self.__class__.model_fields[key].alias
```
loriab, line 633: hooray, I'm glad this custom logic is gone

loriab:
Notice of https://github.com/bennybp/QCElemental/pull/1

loriab (approved):

PR #410, opened 2026-07-30
Add converters between Schrodinger MAE and QCSchema Molecule

TyBalduf:
## Description
Uses schrodinger.structure API to read and write structure objects to convert to/from QCSchema Molecule. 

Mae files can contain essentially arbitrary metadata about a molecule and its atoms, including calculation results. For this converter to Molecule, we just focus on porting structural properties (charge, multiplicity, bond orders), avoiding dumping a bunch of not necessarily structural information in `extras`

Dummies and counterpoise/ghost atoms are accounted for. Maestro limits bond orders to just {0,1,2,3} (aromaticity is inferred from bond patterns); by default we block lossy conversion of non-integer bond orders from Molecule, but we offer a permissive mode in the converter to just round to the nearest valid bond order.

The portions of the Python API used for interconversion are available even if the user has just installed Maestro (which if offered free to academics) rather than the whole Schrodinger Suite.

## Changelog description
Add converters between Schrodinger MAE and QCSchema Molecule

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcelemental/models/v2/molecule.py
+++ b/qcelemental/models/v2/molecule.py
@@ -18,6 +18,8 @@ from numpy.typing import NDArray
 from pydantic import Field, SerializerFunctionWrapHandler, field_validator, model_serializer
 from typing_extensions import Annotated
 
+from qcelemental.molparse import mae
+
 # molparse imports separated b/c https://github.com/python/mypy/issues/7203
 from ...models import QCEL_V1V2_SHIM_CODE
 from ...molparse.from_arrays import from_arrays
@@ -52,6 +54,8 @@ _extension_map = {
     ".psimol": "psi4",
     ".psi4": "psi4",
     ".msgpack": "msgpack-ext",
+    ".mae": "mae",
+    ".maegz": "mae",
 }
 
 
@@ -928,20 +932,20 @@ class Molecule(ProtoModel):
     @classmethod
     def from_data(
         cls,
-        data: Union[str, Dict[str, Any], np.ndarray, bytes],
-        dtype: Optional[str] = None,
+        data: str | dict[str, Any] | np.ndarray | bytes,
+        dtype: str | None = None,
         *,
         orient: bool = False,
-        validate: bool = None,
-        **kwargs: Dict[str, Any],
-    ) -> "Molecule":
+        validate: bool | None = None,
+        **kwargs: Any,
+    ) -> Molecule:
         r"""
         Constructs a molecule object from a data structure.
 
         Parameters
         ----------
         data
-            Data to construct Molecule from
+            Data from which to construct a Molecule.
         dtype
             How to interpret the data, if not passed attempts to discover this based on input type.
         orient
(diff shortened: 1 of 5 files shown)


loriab (commented):
This looks good, thanks! Would you want to hook `from_mae` up to `models.v2.Molecule.from_data()`? It (and `Molecule()`) are the two most common ways of constructing a mol.

I'm curious how common dummy centers (no electrons, no basis functions) are for Jaguar/Maestro? I ask because we've considered them as neglectable for a Cartesian-only format for QCSchema. Psi4 uses them of course (mostly for z-matrix input) but not through schema input. Dummies and embedding charges are a grey area of what belongs in Molecule, so I'd be interested in how necessary you regard them.

TyBalduf:
> This looks good, thanks! Would you want to hook `from_mae` up to `models.v2.Molecule.from_data()`? It (and `Molecule()`) are the two most common ways of constructing a mol.
> 
> I'm curious how common dummy centers (no electrons, no basis functions) are for Jaguar/Maestro? I ask because we've considered them as neglectable for a Cartesian-only format for QCSchema. Psi4 uses them of course (mostly for z-matrix input) but not through schema input. Dummies and embedding charges are a grey area of what belongs in Molecule, so I'd be interested in how necessary you regard them.


I'll try to wire `from_mae` into `from_data` to make it more idiomatic for users to generate a Molecular from a .mae
---
Dummies aren't necessarily crucial and I can see why you might want to exclude them from the molecule schema, as they typically serve as geometric constraints or placeholders to collect output properties like the ESP or density values, rather than meaningful parts of the structure per se.

The main use case where it might be useful on the molecule itself is as a workaround for describing hapticity in transition metal complexes. In systems like ferrocene, we add bonds to dummies as intermediaries, where they form single order bonds to the metal and zero order bonds to the ligand, so we can represent the idea of the metal being singly-bound to the whole ligand, rather than forming individual bonds to the ligand and potentially tripping checks for over-coordination.

So they aren't strictly necessary in terms of making a calculation run or interpreting the output, we can always just strip the dummies and their associated bonds from the output and parse any results stored on them like ESP values into an AtomicResults field. But bonding info is potentially useful for if/when these molecules get passed to visualization software like Maestro or Avogadro, where even with the more complex bonding patterns present with metals, chemists still like to see bonds.

loriab:
Thanks, I think please go ahead and remove dummy from this and #411. A consistent way for handling positions for dummy, hapticity, embedding charges, mid-bond functions, etc. could be devised. I'll post an issue on it.

TyBalduf:
> Thanks, I think please go ahead and remove dummy from this and #411. A consistent way for handling positions for dummy, hapticity, embedding charges, mid-bond functions, etc. could be devised. I'll post an issue on it.

Update this review, #411, and https://github.com/MolSSI/QCEngine/pull/513 to remove any special handling of dummy atoms.

loriab (approved):
Looks good, thank you!

PR #411, opened 2026-07-31
Add a to_string path for Jaguar

TyBalduf:
## Description
Incidentally, updated nucleus reconciliation to accept dummy atoms. These have no atomic mass, charge, or associated basis functions and are just used to define particular internal coordinates for constraints/scans. 

Also updated type hints in touched files to use builtin types, rather than needing to import container types from the typing module. This is supported as of Python 3.10, which as of https://github.com/MolSSI/QCElemental/commit/a8bd79275708764dde0a9b8e5b20454366c6581c is the minimal supported Python version.

## Changelog description
Support generating a template Jaguar input via `to_string`

## Status

- [x] Code base linted
- [x] Ready to go


TyBalduf:
As with https://github.com/MolSSI/QCElemental/pull/410, I implemented some amount of dummy support here as well. But I can roll this back if the preference is to keep this out of Molecule.

loriab (approved):
lgtm, thank you!

PR #415, opened 2026-09-11
Fix serialization error for v1v2 AtomicResult

bennybp:
## Description

When using v1v2 models with the "properties" driver, there is a serialization error due to the pydantic type of `return_result` not handling dictionaries containing numpy arrays. This adds the type annotation so that it gets serialized properly.

## Status

- [X] Code base linted
- [X] Ready to go


j-wags (commented):
I won't approve since I'm not a maintainer, but just chiming in to say that locally installing this change fixes the last remaining CI failure I'm seeing for the openff-qcsubmit update to qcportal 0.70 https://github.com/openforcefield/openff-qcsubmit/pull/398

loriab (approved):

j-wags:
Thanks, @loriab!

